PR #5, opened 2016-11-11
Added texteditor for field labels

### LINKED ISSUE #2: Use a textarea for field titles
So we can use a short title in notification mail and the (sometimes) long text in the label of the checkbox.

rs2487:
fixes #2


Resources/views/forms/fields/default-field.html.twig
```diff
@@ -46,53 +40,62 @@
...
+                               title="width">
+                            <%= translate('sulu_form.width') %>
+                        </label>
+                    </div>
+                {% endblock %}
             </div>
 
             <div class="block block-{{ type }} grid-row small">
-                {% block title_width %}
+                {% block title_default %}
```
alexander-schranz: This block does not contain the `default` block. Because of the issue #3 I would call it `title_short_title` .

Resources/public/js/components/forms/form/main.js
```diff
@@ -92,8 +92,8 @@ define(function () {
          * Binds sandbox and dom events for active tab
          */
         bindFormEvents: function () {
-            var formSelector = this.getActiveFormSelector();
```
alexander-schranz: there should be no empty line at the beginning of a function.

Resources/views/forms/fields/default-field.html.twig
```diff
@@ -46,53 +40,62 @@
...
+                                              class="form-element vertical hidden husky-validate preview-update"
+                                              data-aura-instance-name="title<~=index~>"
+                                              data-type="textEditor"
+                                              data-type-instance-name="title<~=index~>"
+                                              data-mapper-property="title"
+                                              data-form="true"
+                                              data-aura-component="ckeditor@husky"
+                                              data-aura-table="false"
+                                              data-aura-height="100"
+                                              data-aura-max-height="100"
```
alexander-schranz: For usability I think we should remove the max height.

alexander-schranz:
@rs2487 when creating branches you should also add a name to it e.g. `feature/2-texteditor-label`.

Resources/views/forms/fields/types/headline.html.twig
```diff
@@ -3,7 +3,7 @@
 {% block required %}{% endblock %}
 
 {% block title %}
```
alexander-schranz: this block can be removed because it can also be a textarea with links!

Resources/views/forms/fields/types/spacer.html.twig
```diff
@@ -3,7 +3,7 @@
 {% block required %}{% endblock %}
 
 {% block title %}
```
alexander-schranz: Not sure why a `spacer` should have title?

CHANGELOG.md
```diff
@@ -4,3 +4,4 @@
  - ENHANCEMENT #-     Updated namespaces and rename table names
  - ENHANCEMENT #-     Forked from `alexander-schranz/sulu-form-bundle` from commit `36a7cd11562ed0c9f64752b37707cc2771e0baca`
 
+- FEATURE #2 Changed title field for labels to texteditor
```
alexander-schranz: add this before line 4 and intend the text and list the same as the enhancement

PR #55, opened 2017-02-08
Fixed dynamic list factory with hidden fields

dominikmatt:
`$HIDDEN_TYPES` is used in `Sulu\Bundle\FormBundle\ListBuilder\DynamicListFactory`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## dev-develop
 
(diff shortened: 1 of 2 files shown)


alexander-schranz:
@dominikmatt could you add a line to the CHANGELOG.md:

```
 - BUGFIX      #55    Fixed dynamic list build factory with hidden fields
```

PR #68, opened 2017-04-14
Issue #27: make submit button text editable

codebarista:


codebarista:
Now optimized for the StyleCI analysis

PR #71, opened 2017-04-20
Feature #27: make submit button text editable

codebarista:


alexander-schranz:
@codebarista thank you for your contribution and sorry that it took so long!

PR #62, opened 2017-02-17
WIP: Split form builder form structure

dominikmatt:
|Q|A|
|-----|-----|
| Related issues/PRs | #24 |

**TODO:**
- [ ] Remvove structure and implement Collection-Title Provider.
- [ ] Implement Formular-Tab in some Modules.
- [ ] Extend Documentation.

PR #80, opened 2017-06-28
Added type name to dynamic entity and cleanup title provider

alexander-schranz:


Entity/Dynamic.php
```diff
@@ -201,27 +206,33 @@ class Dynamic implements AuditableInterface
...
      * @param string $typeId
      * @param string $locale
      * @param Form $formId
-     * @param null|string $webspaceKey
      * @param array $data
+     * @param string $webspaceKey
+     * @param string $typeName
      */
-    public function __construct($type, $typeId, $locale, $formId, $webspaceKey = null, $data = [])
+    public function __construct($type, $typeId, $locale, $formId, $data = [], $webspaceKey = null, $typeName = '')
```
wachterjohannes: BC-Break?

alexander-schranz: yes but was before since #62

PR #90, opened 2017-07-18
Fixed form data tab and csv export

alexander-schranz:
## 0.2.2

 - BUGFIX      #90    Fixed form data tab pagination.
 - BUGFIX      #90    Fixed form data csv export.
 - ENHANCEMENT #90    Use full width in form data tab.
 - BUGFIX      #90    Fixed back to list in form data tab.
 - BUGFIX      #90    Fixed translation for form data tab name


matt-halliday:
I was going to create an issue for the back to list button! Also I think there is a missing en translation for the tab name... Showing as 'Daten' for en user.

alexander-schranz:
@matt-halliday thank you for reporting did also fix the translation.

PR #96, opened 2017-08-22
Fixed permission problem for template file

alexander-schranz:
@Quehnie can you test this branch?


Quehnie:
@alexander-schranz Tested on the 1.6.3 and the 1.6.2 works nice :) Thank you a lot.

PR #99, opened 2017-09-29
Removed unused function "getReceiverTypes"

wachterjohannes:
Without configuration (in the installation process) the `cache:clear` fails with the exception:

```
PHP Fatal error:  Class Sulu\Bundle\FormBundle\Mail\NullHelper contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (Sulu\Bundle\FormBundle\Mail\HelperInterface::getReceiverTypes) in /app/vendor/sulu/sulu-form-bundle/Mail/NullHelper.php on line 17
```

see the slack channel https://sulu-io.slack.com/archives/C0430GGCV/p1506653176000156

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Changelog
 
-## 0.3.0 (unreleased)
+## 0.3.1
+
+ - HOTFIX      #99    Removed unused function "getReceiverTypes"
+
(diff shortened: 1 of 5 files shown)


alexander-schranz:
@wachterjohannes Add Changelog and add to UPGRADE that this function has been removed.

PR #101, opened 2017-10-18
Remove null value

codebarista:
If only the email was set, the function always returned "null".

### DIFF
--- a/Configuration/FormConfigurationFactory.php
+++ b/Configuration/FormConfigurationFactory.php
@@ -345,6 +345,10 @@ class FormConfigurationFactory
      */
     private function getEmail($email, $name = null)
     {
+        if (!$email) {
+            return null;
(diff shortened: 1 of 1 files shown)


Configuration/FormConfigurationFactory.php
```diff
@@ -345,10 +345,6 @@ private function getEmailFromDynamic(Dynamic $dynamic)
      */
     private function getEmail($email, $name = null)
     {
-        if (!$email || !$name) {
```
alexander-schranz: the `!$email` check should still exist else the fallback mechanism for from and to doesn't work.

```
        if (!$email) {
            return null;
        }
```

alexander-schranz: @codebarista did correct the check

PR #107, opened 2017-11-03
Add title automatically to new added block

alexander-schranz:


Resources/views/forms/fields/default-field.html.twig
```diff
@@ -94,7 +94,7 @@
                                               data-aura-webspace=""
                                               data-aura-locale="<%= options.language %>"
                                               data-validation-required="true"
-                                              data-aura-enter-mode="br"></textarea>
+                                              data-aura-enter-mode="br"><%= translate('{{ type.configuration.title }}') %></textarea>
```
alexander-schranz: @wachterjohannes I'm not sure how I can get the correct translation of the content language instead of the users current locale. The `renderTemplate` seems to have no possibility to add a querystring and the frontend translation only have the current locale available.

wachterjohannes: could use symfony translations here? and if you want that you have to omit the `renderTemplate` and implement the loading on your own.

PR #135, opened 2018-01-24
Fix page crash when form was removed

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Catch exception when a form is loaded which not longer exist.

#### Why?

That delete forms are not throwing exceptions on website page when the formId is still referenced in PHPCR.

#### Example Usage

1. Create a form
2. Create a page with a form linked
3. Delete the form
4. Visit the page

#### To Do

- [x] Update CHANGELOG.md


wachterjohannes:
@alexander-schranz the tests are failing

PR #136, opened 2018-01-24
Fix default value for choice types

### LINKED ISSUE #133: Radio button field (required) not selected by default.
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | yes
| Bundle Version | All
| Sulu Version | All
| Browser Version | All

#### Actual Behavior

If you create a radio button group which is `required` there is no radio button preselected.

#### Expected Behavior

The first radio button should be selected by default, if you set the field as required.

#### Possible Solutions

Set the first radio button as selected, if the radio button field is set to required.

### LINKED ISSUE #134: Radio button default value field.
| Q | A
| --- | ---
| Bug? | no
| New Feature? | yes
| Bundle Version | All
| Sulu Version | All
| Browser Version | All

#### Actual Behavior

There is no way to set a default selection of the radio button.

#### Expected Behavior

A possibility to set one of the radio button values as default value.

#### Possible Solutions

Add a text-field where you can set the default value of the radio buttons.

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #134 fixes #133
| License | MIT

#### What's in this PR?

Add defaultValue to all choice form types and set placeholder correctly when set.

#### Why?

See linked issues.

- [x] Update CHANGELOG.md


Resources/config/doctrine/FormFieldTranslation.orm.xml
```diff
@@ -10,7 +10,7 @@
 
         <field name="title" column="title" type="text" nullable="true"/>
         <field name="placeholder" column="placeholder" type="string" length="255" nullable="true"/>
-        <field name="defaultValue" column="defaultValue" type="string" length="255" nullable="true"/>
+        <field name="defaultValue" column="defaultValue" type="text" nullable="true"/>
```
wachterjohannes: upgrade?

Dynamic/Types/MultiChoiceTrait.php
```diff
@@ -22,18 +22,28 @@
...
+        $multiple = false
+    ) {
         $options = [];
         if ($translation) {
             // Placeholder.
             $options['placeholder'] = $translation->getPlaceholder();
 
+            if (!$options['placeholder'] && !$multiple && !$required) {
+                $options['placeholder'] = 'sulu_form.no_choice';
+                $options['translation_domain'] = 'messages';
```
wachterjohannes: why does this options depends on multiple and required?

Entity/FormFieldTranslation.php
```diff
@@ -129,7 +129,7 @@ public function getField()
      */
     public function getPlaceholder()
```
wachterjohannes: phpdoc is now wrong - should be `string|boolean` which seems to be wrong in my opinion. why not returning null if the placeholder is not set?

Dynamic/Types/MultiChoiceTrait.php
```diff
@@ -22,18 +22,28 @@
...
-    public function getChoiceOptions(FormFieldTranslation $translation, $expanded = false, $multiple = false)
-    {
+    private function getChoiceOptions(
+        FormFieldTranslation $translation,
+        $required = false,
+        $expanded = false,
+        $multiple = false
+    ) {
         $options = [];
         if ($translation) {
```
wachterjohannes: is it possible that this is null? I think the typehint impeded that.

Dynamic/Types/MultiChoiceTrait.php
alexander-schranz: `multicheckbox` and `multiselect` does not render the placeholder only Radio Button or Single Select will render the placeholder as empty field and this should only happen when not required.

Entity/FormFieldTranslation.php
alexander-schranz: there is a check for `false` and only when its really false the placeholder is not rendered.

Resources/config/doctrine/FormFieldTranslation.orm.xml
alexander-schranz: added

PR #155, opened 2018-04-12
Fix controller in example template

Quehnie:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fix controller in example template.

#### Why?

False controller in example.


alexander-schranz:
moved to #156

PR #165, opened 2018-05-02
Fix rendering of checkbox labels that contain links to external or internal pages

wimvds:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR fixes the rendering of checkbox labels that contain links to external or internal pages.

#### Why?

Currently the HTML markup is dumped instead of the link (due to Twig escaping the output).

### DIFF
--- a/Resources/views/themes/dynamic.html.twig
+++ b/Resources/views/themes/dynamic.html.twig
@@ -103,3 +103,29 @@
         {%- endif -%}
     {%- endfor -%}
 {%- endblock attributes -%}
+
+{# Make sure links also work in labels #}
+{% block checkbox_radio_label -%}
+    {%- if widget is defined -%}
+        {%- if required -%}
+            {%- set label_attr = label_attr|merge({class: (label_attr.class|default('') ~ ' required')|trim}) -%}
+        {%- endif -%}
+        {%- if parent_label_class is defined -%}
+            {%- set label_attr = label_attr|merge({class: (label_attr.class|default('') ~ ' ' ~ parent_label_class)|trim}) -%}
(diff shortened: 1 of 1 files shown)


Resources/views/themes/dynamic.html.twig
```diff
@@ -103,3 +103,29 @@
         {%- endif -%}
     {%- endfor -%}
 {%- endblock attributes -%}
+
+{# Make sure links also work in labels #}
+{% block checkbox_radio_label -%}
```
alexander-schranz: This block can't be found in https://github.com/symfony/symfony/blob/3.4/src/Symfony/Bridge/Twig/Resources/views/Form/form_div_layout.html.twig which version of symfony and the sulu form bundle did you use? Also the variable `widget` is strange here.

The output of correct HTML for labels is fixed in: https://github.com/sulu/SuluFormBundle/blob/ff5fe44ef9c4fd34a8e0d746082a6b8a1b289645/Resources/views/themes/dynamic.html.twig#L18-L24

Did you accidently overwrite that part in your project?

wimvds: My bad, just noticed our form theme is using Bootstrap 3 (https://github.com/symfony/symfony/blob/8659ba9f1f7468ff98fc3af0bd5185e4d207c8d2/src/Symfony/Bridge/Twig/Resources/views/Form/bootstrap_3_layout.html.twig#L82-L105). Guess we'll stick with overriding the default theme...

Issue #181, opened 2019-01-10
Compatibility issue with doctrine/orm 2.6


MichelFalke:
| Q | A
| --- | ---
| Bug? | Yes
| New Feature? | no
| Bundle Version | all versions
| Sulu Version | all versions
| Browser Version | all versions

#### Actual Behavior

Doctrine ORM EntityRepository has added the function count. Since the FormRepository extends from it and has it's own count with multiple parameters this gives a conflict.

#### Expected Behavior

Don't conflict on inheritance

#### Steps to Reproduce

Update doctrine/orm to ^2.6

#### Possible Solutions

Rename the methods in the FormRepository
count becomes countByFilters
findAll becomes loadAll
findById becomes loadById

alexander-schranz:
#182

PR #182, opened 2019-01-11
Fix compatibility with doctrine/orm ^2.6 (#181)

MichelFalke:
* Renamed functions because of an inheritance issue with doctrine ^2.6


UPGRADE.md
```diff
@@ -1,5 +1,11 @@
 # Upgrade
 
+## 1.0.0-RC6
+For the compatibility with `doctrine/orm ^2.6` the function names of the FromRepository are edited to avoid inheritance issues.
```
alexander-schranz: add an empty line before and after this line. and there is a small typo in `Form...` (From...)

Issue #190, opened 2019-07-04
Upgrade Roadmap to Sulu 2.0 compatibility


alexander-schranz:
| Q | A
| --- | ---
| Bug? | no
| New Feature? | yes
| Bundle Version | for 2.0
| Sulu Version | for 2.0
| Browser Version | Browser name and version

## Changes

- [x] **1. Implement new List of Forms** (implemented in #190)

Remove fieldsAction and create list xml for forms.

- [x] **2. Implement form for forms** (implemented in #190)

Create form without fields section. 

- [x] **3. Extend FormMetadataProvider (Sulu Core changes needed)** (implemented in #192)

~~EvaluateFormItemExpressions should be an own service?
MetadataProviderRegistry should handle Providers as chain so an own Provider can be implemented to add the FormMetaData for the Fields blocks.~~

Extend sulu to have multiple form meta data loaders and create a custom one in the form bundle to load the form_details form.

- [x] **4. Dynamic List Fields implementation for Data Tab (Sulu Core changes needed)** (implemented in #197)

We need a option in the RouteBuilder which does made it possible to add parameters to the metadata request. So the ListMetaProvider chain get this options and can dynamically add FieldDescriptors. 

Use addRouterAttributesToListMetadata ... ['id' => 'id']
Use addRouterAttributesToListStore to add a filter for this list. 

- [x] **5. Page Tab** (#211) 

Register a new route for Pages and set the tabCondition dynamically over loadedStructureMetadata to add the tab only for this structures which has the single_form_selection content type.

Use addRouterAttributesToListStore to add a filter for this list. 

- [x] **6. Database**

Currently there is a strange mix about data when having a look at Dynamic.orm.xml

https://github.com/sulu/SuluFormBundle/blob/1.0.0-RC7/Resources/config/doctrine/Dynamic.orm.xml

The entity should be reduced to some basic fields:

 - id: int
 - (uuid: string)
 - type (maybe resourceKey): string
 - typeId (maybe resourceId): string
 - locale: ?string (5)
 - webspaceKey: ?string (64)
 - typeName (maybe resourceName / resourceTitle): ?string (128)
 - data: array
 - auditableFields
 - form: relation to form

For this all other fields need to be migrated into data field json, fields should only be set in data when the column have data for it.

This will also fix the MySQL 8 Incompatibility see https://github.com/sulu/SuluFormBundle/pull/199

- [x] **7. Dynamic.php** #255 

[...]

PR #191, opened 2019-07-11
Upgrade Sulu 2.0

markusmoosbrugger:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | 
| Deprecations? | no
| License | MIT

#### What's in this PR?

- Implement new list of forms
- Implement form for form details (form fields are static at the moment)
- Export single_form_selection content type

#### Why?

Upgrade to work with Sulu 2.0

#### BC Breaks/Deprecations

The content type **single_form_selection** has to be used in a different way than the content type **form_select** previously. The param name "type" has changed to "resourceKey" because "type" is overwritten internally to specify the type of the content type (single_select in this case) itself.

Example Usage:
```xml
<property name="form" type="single_form_selection">
     <meta>
         <title lang="de">Formular</title>
         <title lang="en">Form</title>
     </meta>
     <params>
         <param name="resourceKey" value="page"/>
      </params>
</property>
```

#### To Do

- [ ] Update CHANGELOG.md
- [ ] Create documentation


composer.json
```diff
@@ -3,16 +3,16 @@
     "description": "Bundle for creating forms in Sulu.",
     "type": "sulu-bundle",
     "require": {
-        "symfony/symfony": "^2.8.7 || ^3.0",
-        "symfony/swiftmailer-bundle": "^2.3 || ^3.0",
+        "symfony/symfony": "^2.8.7 || ^3.0 || ^4.0",
```
alexander-schranz: we should not longer require  the whole `symfony/symfony` framework here we should replace this with:

```
        "symfony/config": "^3.4 || ^4.0",
        "symfony/dependency-injection": "^3.4 || ^4.0",
        "symfony/event-dispatcher": "^3.4 || ^4.0",
        "symfony/form": "^3.4 || ^4.0",
        "symfony/validator": "^3.4 || ^4.0",
```

composer.json
```diff
@@ -3,16 +3,16 @@
...
+        "symfony/swiftmailer-bundle": "^2.3 || ^3.0 || ^4.0",
```
alexander-schranz: it seems the swiftmailer is versionated not the same way as symfony so we need remove ^4.0

```
        "symfony/swiftmailer-bundle": "^2.3 || ^3.0",
```

Admin/FormAdmin.php
```diff
@@ -75,9 +156,15 @@ public function getSecurityContexts()
         return [
             'Sulu' => [
                 'Form' => [
-                    'sulu.form.forms',
+                    'sulu.form.forms' => [
```
alexander-schranz: /cc @chirimoya this seems to be a BC Break that it need that permissions need to be defined

Resources/config/form-fields/header.xml
```diff
@@ -0,0 +1,69 @@
...
+        </meta>
+        <params>
+            <param name="label">
+                <meta>
+                    <title>sulu_form.width</title>
+                </meta>
+            </param>
+            <param name="default_value" value="0"/>
+            <param name="values" type="collection">
+                <param name="0">
```
kleinkoerkamp: Name should be 'full' instead of '0' so that's compatible with Form/Type/DynamicFormType.php

markusmoosbrugger: Fixed. Thanks for pointing this out.

Dynamic/Types/RecaptchaType.php
```diff
@@ -30,7 +30,7 @@ public function getConfiguration()
     {
         return new FormFieldTypeConfiguration(
             'sulu_form.type.recaptcha',
-            'SuluFormBundle:forms:fields/types/recaptcha.html.twig',
+            'SuluFormBundle/Resources/config/form-fields/default_field.xm',
```
alexander-schranz: .xml

Dynamic/Types/MailchimpType.php
```diff
@@ -42,7 +42,7 @@ public function getConfiguration()
     {
         return new FormFieldTypeConfiguration(
             'sulu_form.type.mailchimp',
-            'SuluFormBundle:forms:fields/types/mailchimp.html.twig',
+            'SuluFormBundle/Resources/config/form-fields/field_mailchimp.xm',
```
alexander-schranz: .xml

Resources/config/services.xml
```diff
@@ -66,21 +58,21 @@
...
-                 class="Sulu\Bundle\FormBundle\Content\Types\FormSelect">
+        <service id="sulu_form.content_type.single_form_selection"
+                 class="Sulu\Bundle\FormBundle\Content\Types\SingleFormSelection">
             <argument>SuluFormBundle:Template:content-types/form_select_simple.html.twig</argument>
             <argument type="service" id="sulu_form.repository.form" />
             <argument type="service" id="sulu_form.builder" />
             <argument type="service" id="sulu_form.reference_store.form"/>
 
-            <tag name="sulu.content.type" alias="form_select"/>
+            <tag name="sulu.content.type" alias="single_form_selection"/>
```
alexander-schranz: Add the following to the UPGRADE.md:

~~~~
# dev-develop

The content type for form selection has been changed from `form_select` to `single_form_selection` also the param `type` has changed to `resourceKey`:

**Before**

```xml
<property name="form" type="form_select">
    <params>
        <param name="type" value="page" />
    </params>
</property>
```

**Before**

```xml
<property name="form" type="single_form_selection">
    <params>
        <param name="resourceKey" value="page" />
    </params>
</property>
```
~~~~

alexander-schranz:
We will merge this state into develop branch to keep the PR small as possible.

PR #197, opened 2019-08-08
Upgrade to Sulu 2.0 - Dynamic List

markusmoosbrugger:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | https://github.com/sulu/sulu/pull/4671
| License | MIT

#### What's in this PR?

This PR adds a ListMetadataLoader which is responsible for loading the dynamic list metadata for the data tab.

The **DynamicListMetadataLoader** implements the **ListMetadataLoaderInterface** and has been tagged in the services.xml file as **sulu_admin.list_metadata_loader** in order to get injected into the **FormMetadataProvider** in Sulu Core. For this functionality also Sulu Core changes are required (see PR https://github.com/sulu/sulu/pull/4671).

#### Why?

The data tab in the Form Bundle should display all submitted data of the currently selected form. As there are more than 20 different form types in total, only the types which are included in the current form should be displayed as rows in the data table. Therefore a dynamic list metadata loader is needed.

#### BC Breaks/Deprecations

#### To Do

- [ ] Update CHANGELOG.md
- [ ] Create documentation


Admin/FormAdmin.php
```diff
@@ -26,6 +26,7 @@
 class FormAdmin extends Admin
 {
     const LIST_ROUTE = 'sulu_form.list';
+    const LIST_ROUTE_DATA = 'sulu_form.list.data';
```
danrot: The dot here is acting kind like a namespace. So I would name it `sulu_form.edit_form.data`, because it is only appearing in the edit form.

Admin/FormAdmin.php
```diff
@@ -142,6 +146,16 @@ function (Localization $localization) {
                 ->addToolbarActions($formToolbarActions)
                 ->setParent(static::EDIT_FORM_ROUTE)
                 ->getRoute(),
+            $this->routeBuilderFactory->createListRouteBuilder(static::LIST_ROUTE_DATA, '/form_data')
```
danrot: Would use just `data` instead of `form_data`, since `form` is already part of the URL

Admin/FormAdmin.php
```diff
@@ -142,6 +146,16 @@ function (Localization $localization) {
...
+                ->setResourceKey('dynamics')
```
danrot: Maybe `dynamic_forms`? If you have a lot of resource keys just `dynamics` does not tell a lot.

Controller/DynamicController.php
```diff
@@ -11,23 +11,85 @@
...
+    private $viewHandler;
+
+    /**
+     * DynamicController constructor.
+     * @param DynamicRepository $dynamicRepository
+     * @param DynamicListFactory $dynamicListFactory
+     * @param MediaManager $mediaManager
+     * @param EntityManager $entityManager
+     * @param FormRepository $formRepository
+     * @param ViewHandler $viewHandler
```
danrot: Remove that PHPDoc

Controller/DynamicController.php
```diff
@@ -93,11 +152,11 @@ public function cgetFieldsAction(Request $request)
...
         if ($form) {
             $locale = $this->getLocale($request);
-            $fieldDescriptors = $this->get('sulu_form.list_builder.dynamic_list_factory')
+            $fieldDescriptors = $this->dynamicListFactory
                 ->getFieldDescriptors($form, $locale);
         }
 
-        return $this->handleView($this->view(array_values($fieldDescriptors)));
+        return $this->viewHandler->handle($this->view(array_values($fieldDescriptors)));
     }
```
danrot: I think you don't need this function at all anymore, so it can be safely deleted. This was the way the metadata for lists was loaded in the old UI.

Metadata/DynamicListMetadataLoader.php
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    private function getForm(array $metadataOptions, string $locale): ?Form
+    {
+        if (!array_key_exists('id', $metadataOptions)) {
+            return null;
+        }
+        $entity = $this->formManager->findById($metadataOptions['id'], $locale);
+
+        if (!$entity) {
```
danrot: Isn't that already null when `findById` doesn't find anything? Than this `if` could be omitted.

Resources/config/routing_api.yml
```diff
@@ -13,9 +13,11 @@ sulu_form.forms:
 
 sulu_form.dynamic:
     type: rest
-    resource: Sulu\Bundle\FormBundle\Controller\DynamicController
+    resource: sulu_form.dynamic_controller
     prefix: form
     name_prefix: sulu_form.
+    options:
+        expose: true
```
danrot: What is this doing? :thinking:

Resources/config/services.xml
```diff
@@ -115,7 +115,7 @@
 
         <!-- List Builder -->
         <service id="sulu_form.list_builder.dynamic_list_factory"
-                 class="Sulu\Bundle\FormBundle\ListBuilder\DynamicListFactory">
+                 class="Sulu\Bundle\FormBundle\ListBuilder\DynamicListFactory" public="true">
```
danrot: Do you still need these services to be public? Please check all services in this file.

Resources/translations/admin.de.json
```diff
@@ -18,6 +18,7 @@
   "sulu_form.deactivate_notify": "Benachrichtungsemails deaktivieren",
   "sulu_form.deactivate_success": "Bestätigungsemails deaktivieren",
   "sulu_form.receivers": "Empfänger",
+  "sulu_form.data": "Data",
```
danrot: The below translation seems to be the correct one

Resources/config/routing_api.yml
markusmoosbrugger: As far as I understand it, this is needed to make the routes for the `DynamicController` available. 
If I remove this options, I get an error that the route doesn't exist.

danrot: Ah, ok, so this is probably from the `FOSJsRoutingBundle`... I think in order to keep stuff consistent, I would choose the same way as it was done using the [`sulu_admin.metadata` route](https://github.com/sulu/sulu/pull/4671/files#diff-552d74d7863d5ff3781de14203825633R102).

Resources/config/services.xml
markusmoosbrugger: - `DynamicListMetadataLoader` and `DynamicFormMetadataLoader`: removed public and defined services for tests in `config.yml`
- `DynamicController`: needs to be public
- `DynamicRepository`: removed public
- `DynamicListFactory`: removed public
- `FormManager`: needs to be public

PR #205, opened 2019-10-06
Fix accidentally added item in composer

reyostallenberg:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT

#### What's in this PR?
A fix of an accidentally added item in composer.json

#### Why?
Makes composer.json valid

#### Example Usage
-

#### BC Breaks/Deprecations
-

- 

#### To Do
-

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "doctrine/data-fixtures": "^1.0",
         "doctrine/doctrine-bundle": "^1.4",
         "drewm/mailchimp-api": "^2.2",
-        "excelwebzone/recaptcha-bundle": "^1.4.2"
+        "excelwebzone/recaptcha-bundle": "^1.4.2",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -15,7 +15,6 @@
         "symfony/translation": "^3.4 || ^4.0",
         "twig/twig": "^1.26 || ^2.0"
     },
-    "excelwebzone/recaptcha-bundle": "^1.4.2",
```
alexander-schranz: This should be a test dependency.

reyostallenberg: Okey, fixed that

alexander-schranz: Thank you!

alexander-schranz:
@reyostallenberg Thank you!

PR #216, opened 2019-12-18
#194 Add sender the same as parameter for Sender-header on envelope

### LINKED ISSUE #194: Add Sender to Mail
| Q | A
| --- | ---
| Bug? | no
| New Feature? | yes
| Bundle Version | 1.0.0

On some servers the mails send from Sulu forms end up in spam-folders because the 'envelope-from' header is missing in the message and is auto-generated on the server when sending a e-mail.

To fix this the sender must be added to the Swift_Message with setSender() in Mail/Helper.php. This can be the same as the from-address or maby make it configurable with a fallback to the from-address.

If I can find the time I will make a PR for it. At least if there isn't a reason why not to add it ;-)

lkeijmel:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #194 

#### What's in this PR?

Add extra "sender" configuration parameter to add a Sender-header to the mail envelope.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -52,7 +52,7 @@ class Configuration implements ConfigurationInterface
                 ->children()
                     ->scalarNode('from')->defaultValue('%sulu_admin.email%')->end()
                     ->scalarNode('to')->defaultValue('%sulu_admin.email%')->end()
-                    ->scalarNode('sender')->defaultValue('%sulu_admin.email%')->end()
+                    ->scalarNode('sender')->defaultValue(null)->end()
                     ->arrayNode('templates')
                         ->addDefaultsIfNotSet()
                         ->children()
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -52,6 +52,7 @@ class Configuration implements ConfigurationInterface
                 ->children()
                     ->scalarNode('from')->defaultValue('%sulu_admin.email%')->end()
                     ->scalarNode('to')->defaultValue('%sulu_admin.email%')->end()
+                    ->scalarNode('sender')->defaultValue('%sulu_admin.email%')->end()
                     ->arrayNode('templates')
                         ->addDefaultsIfNotSet()
                         ->children()
--- a/Mail/Helper.php
+++ b/Mail/Helper.php
@@ -41,10 +41,10 @@ class Helper implements HelperInterface
     protected $logger;
 
     /**
-     * @param \Swift_Mailer   $mailer
-     * @param string          $fromMail
-     * @param string          $toMail
-     * @param null|string     $sender
+     * @param \Swift_Mailer $mailer
+     * @param string $fromMail
+     * @param string $toMail
+     * @param string|null $sender
      * @param LoggerInterface $logger
      */
     public function __construct(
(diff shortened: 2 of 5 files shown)


alexander-schranz:
@lkeijmel Thank you for your pull request. Need to discuss this with the team. The problem I see is what should be the sender when I enter a from address in the sulu UI.

For me I think most would make sense that `from` is always the configurated one from `sulu_form.yaml` file and `sender` is what the content manager can set as `from`. Because as Sender I understand you could enter all E-Mail addresses not only the one the server is allowed to send as email address.

lkeijmel:
@alexander-schranz no problem to discus it but I think you should swap your argumenta. The `sender` should configured only once as it's server specific and the `from` could be configured by a content manager. Making `sender` also configurable by a content manager sounds also like a good solution but it's easy to confuse the headers resulting that a e-mail send from the form is marked as spam.

A little information about the differences between `from` and `sender`:

**Envelope From (RFC5321.MAIL FROM)**
- Used by the SMTP server to generate NDR
- Used by SPF filter to determine if it came from the designated IP address.

**Header From (RFC5322.FROM)**
- Used by the email client to display information in the From field.
- Used by DMARC filter to confirm if the message is authentic

Source: https://www.xeams.com/difference-envelope-header.htm

Mail/Helper.php
```diff
@@ -90,6 +96,10 @@ public function sendMail(
         $message->setFrom($fromMail);
         $message->setTo($toMail);
 
+        if (null !== $this->sender) {
```
alexander-schranz: use `if ($this->sender) {` so it can not set accidently to an empty string

Mail/Helper.php
```diff
@@ -32,26 +32,32 @@ class Helper implements HelperInterface
...
     /**
      * @var LoggerInterface
      */
     protected $logger;
 
     /**
-     * @param \Swift_Mailer $mailer
-     * @param string $fromMail
-     * @param string $toMail
+     * @param \Swift_Mailer   $mailer
```
alexander-schranz: we don't indent the param names

Mail/Helper.php
```diff
@@ -32,26 +32,32 @@ class Helper implements HelperInterface
      */
     protected $fromMail;
 
+    /** @var null|string */
```
alexander-schranz: change to:

```
/**
 * @var string|null
 */
```

Mail/Helper.php
```diff
@@ -32,6 +32,9 @@ class Helper implements HelperInterface
      */
     protected $fromMail;
 
+    /** @var string|null */
```
alexander-schranz: Please multiline  :)

DependencyInjection/Configuration.php
```diff
@@ -52,6 +52,7 @@ public function getConfigTreeBuilder()
                 ->children()
                     ->scalarNode('from')->defaultValue('%sulu_admin.email%')->end()
                     ->scalarNode('to')->defaultValue('%sulu_admin.email%')->end()
+                    ->scalarNode('sender')->defaultValue('%sulu_admin.email%')->end()
```
alexander-schranz: Sorry didn't see this before. I would set the defaultValue to `null` so its only set when configured by someone.

alexander-schranz:
@lkeijmel Thank you!

PR #235, opened 2020-04-03
Fix rendering of static forms

### LINKED ISSUE #233: Static Form Compatibility Issues with Sulu 2.0
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Bundle Version | develop @ 3ec157a943002242f76f606888ce72b3c1732a8f
| Sulu Version | 2.0.4
| Browser Version | Browser independent

#### Actual Behavior

When trying to display a static form under the new Sulu 2.0 (for example after Upgrading) the form will break with the following Error:
```
The "Sulu\Bundle\FormBundle\Controller\FormWebsiteController::getParameter()" method is missing a parameter bag to work properly. Did you forget to register your controller as a service subscriber? This can be fixed either by using autoconfiguration or by manually wiring a "parameter_bag" in the service locator passed to the controller.
```

#### Expected Behavior

Either the configuration should be fixed, which is currently breaking the static forms, or it would be nice if the Documentation gets adapted with how to quickly change a static form to a Symfony Form.

#### Steps to Reproduce

I configured the bundle as follows:

In contact-form.xml (page template)
```
<controller>Sulu\Bundle\FormBundle\Controller\FormWebsiteController:formAction</controller>
```

In services.yml
```
contact_form:
        class: 'App\Form\Type\ContactRequestFormType'
        arguments:
            - '@translator'
        tags:
            - 'form.type'
```

Configure a Form Type and Entity like in the Documentation.

#### Possible Solutions

A possible solution would be to update the Documentation to also contain a section Upgrade Static forms to Symfony forms.

To fix static forms you will probably need to inject the `parameter_bag` service in `vendor/sulu/sulu-form-bundle/DependencyInjection/SuluFormExtension.php`.

luca-rath:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #233 
| License | MIT

#### What's in this PR?

Fix rendering of static forms

#### Why?

To fix #233

### DIFF
--- a/Configuration/FormConfigurationFactory.php
+++ b/Configuration/FormConfigurationFactory.php
@@ -130,7 +130,6 @@ class FormConfigurationFactory
         $adminMailConfiguration->setReplyTo($type->getNotifyReplyToMailAddress($formData));
         $adminMailConfiguration->setAddAttachments($type->getNotifySendAttachments($formData));
         $adminMailConfiguration->setTemplate($type->getNotifyMail($formData));
-        $adminMailConfiguration->setPlainTextTemplate($this->mailAdminPlainTextTemplate);
         $adminMailConfiguration->setTemplateAttributes($attributes);
 
         return $adminMailConfiguration;
@@ -156,7 +155,6 @@ class FormConfigurationFactory
         $websiteMailConfiguration->setReplyTo($type->getCustomerReplyToMailAddress($formData));
         $websiteMailConfiguration->setAddAttachments(false); // Currently not implemented in the AbstractType.
         $websiteMailConfiguration->setTemplate($type->getCustomerMail($formData));
-        $websiteMailConfiguration->setPlainTextTemplate($this->mailWebsitePlainTextTemplate);
         $websiteMailConfiguration->setTemplateAttributes($attributes);
 
         return $websiteMailConfiguration;
--- a/Configuration/FormConfigurationFactory.php
+++ b/Configuration/FormConfigurationFactory.php
@@ -130,6 +130,7 @@ class FormConfigurationFactory
         $adminMailConfiguration->setReplyTo($type->getNotifyReplyToMailAddress($formData));
         $adminMailConfiguration->setAddAttachments($type->getNotifySendAttachments($formData));
         $adminMailConfiguration->setTemplate($type->getNotifyMail($formData));
+        $adminMailConfiguration->setPlainTextTemplate($this->mailAdminPlainTextTemplate);
         $adminMailConfiguration->setTemplateAttributes($attributes);
 
         return $adminMailConfiguration;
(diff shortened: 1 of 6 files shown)


Resources/config/services.xml
```diff
@@ -223,6 +225,16 @@
             <tag name="sulu.context" context="admin"/>
         </service>
 
+        <!-- Form Website Controller -->
+        <service id="sulu_form.form_website_controller" class="Sulu\Bundle\FormBundle\Controller\FormWebsiteController" public="true">
```
luca-rath, line 229: Can't tag this controller with `sulu.context` `website`, because otherwise i wouldn't be able to alias it. @alexander-schranz wdyt?

alexander-schranz: Do we need to register it as a service as I think the `Sulu\Bundle\WebsiteBundle\Controller\DefaultController` is also no service or? What happens if we remove the service definition?

We should split the controller `Website/FormStaticController` and `Website/FormTokenController`.

luca-rath: Yes it is.
```xml
<service id="sulu_website.default_controller" class="Sulu\Bundle\WebsiteBundle\Controller\DefaultController">
    <tag name="container.service_subscriber" />
    <tag name="controller.service_arguments" />
    <tag name="sulu.context" context="website" />
    <call method="setContainer">
        <argument type="service" id="Psr\Container\ContainerInterface" />
    </call>
</service>
<service id="Sulu\Bundle\WebsiteBundle\Controller\DefaultController" alias="sulu_website.default_controller"/>
```
This is the definition of sulu's `DefaultController`. But I have absolutely no idea why it's working here, that the controller is tagged with `sulu.context` `website`.

luca-rath: If I remove the alias, I get the same error as in #233

luca-rath: Have a look at my latest commit, this seems to solve it. Maybe it's working for the `DefaultController`, because neither the service nor the alias is public. But if I remove the public attribute from the `FormWebsiteController` and it's alias, I also get the same error as in #233.

Configuration/FormConfigurationFactory.php
```diff
@@ -130,6 +130,7 @@ private function buildAdminMailConfigurationByTypeAndData(AbstractType $type, $f
         $adminMailConfiguration->setReplyTo($type->getNotifyReplyToMailAddress($formData));
         $adminMailConfiguration->setAddAttachments($type->getNotifySendAttachments($formData));
         $adminMailConfiguration->setTemplate($type->getNotifyMail($formData));
+        $adminMailConfiguration->setPlainTextTemplate($this->mailAdminPlainTextTemplate);
```
alexander-schranz, line 133: This templates are not for static forms. And should stay null

Configuration/FormConfigurationFactory.php
```diff
@@ -155,6 +156,7 @@ private function buildWebsiteMailConfigurationByTypeAndData(AbstractType $type,
         $websiteMailConfiguration->setReplyTo($type->getCustomerReplyToMailAddress($formData));
         $websiteMailConfiguration->setAddAttachments(false); // Currently not implemented in the AbstractType.
         $websiteMailConfiguration->setTemplate($type->getCustomerMail($formData));
+        $websiteMailConfiguration->setPlainTextTemplate($this->mailWebsitePlainTextTemplate);
```
alexander-schranz, line 159: This templates are not for static forms. And should stay null

Configuration/FormConfigurationFactory.php
luca-rath: But it's beeing used by the `Form\Handler::sendMail()` method.

PR #243, opened 2020-04-24
Fix field order after save

cirykpopeye:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Related issue | #245 
| Deprecations? | no
| License | MIT

#### What's in this PR?

A fix for the sorting issue on save

#### Why?

When you save the form after you sorted some fields, the fields get reset to it's original position until page reload.


alexander-schranz:
@cirykpopeye thank you!

Issue #209, opened 2019-11-28
Not possible to type in "Choices" field while using "Select" block


michael-hagspiel:
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Bundle Version | Specific version or SHA of a commit
| Sulu Version | sulu 2.0.2

#### Actual Behavior

While using a "Select" block it is not possible to type anything into the "Choices" field (see screenshot). This might be only for some languages, as it seems to work in all languages except of german.

![Issue_FormBundle_Select](https://user-images.githubusercontent.com/58299847/69812779-2132f680-11f1-11ea-9c64-67de8b4567a1.png)

#### Expected Behavior

It is possible to type something into the "Choices" field.

danrot:
I think that might be related with sulu/sulu#4501, but I couldn't reproduce that issue myself... If I create a new form, add a select field in english, save it and switch to german, I can still enter something into the choices field. Can you give more detailed steps to reproduce?

michael-hagspiel:
Steps to reproduce:
1. Create a new form with a block, which is not a Select
2. Save the form and go back to the overview page of all forms
3. Open the form again and change the type of the already made block to Select
4. Now you can't type into the Choices Field of that block

alexander-schranz:
Short analyse I think it seem like the problem is that the response of the `options` parameter is `[]` instead of `{}`. When using `<property name="options/choices" type="text_area"/>` in the form.

alexander-schranz:
This is fixed by #244

PR #244, opened 2020-04-24
Fix/radio options cant edit

cirykpopeye:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Related issue | #209
| Deprecations? | no
| License | MIT

#### What's in this PR?

A fix where you could not type inside the choices area of a radiobutton

#### Why?

When the translation does not exist, the options of a radio button return [], this has conflicts with react so when options is empty I now return { choices: [] } for the radio button type only.

### DIFF
--- a/Controller/FormController.php
+++ b/Controller/FormController.php
@@ -366,6 +366,11 @@ class FormController extends AbstractRestController implements ClassResourceInte
             $fields[] = $fieldData;
         }
 
+        // Sort fields with correct order
+        usort($fields, function($fieldA, $fieldB) {
+            return $fieldA['order'] > $fieldB['order'];
+        });
+
         // Api Entity
         return array_merge(
             [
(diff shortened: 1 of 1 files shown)


Controller/FormController.php
```diff
@@ -363,6 +363,10 @@ private function getApiEntity(Form $entity, string $locale): array
                 $fieldData['options'] = $fieldTranslation->getOptions();
             }
 
+            if ($fieldData['type'] === 'radioButtons' && empty($fieldData['options'])) {
```
alexander-schranz: I think this should be genrally be fixed for all types the problem is that `options` does return an array instead of an object I think and so we need todo:

```php
if (empty($fieldData['options'])) {
     $fieldData['options'] = new stdClass(); // convert options to "{}"
}
```

Then this should also work for multi dropdown and multi checkboxes

cirykpopeye: Correct @alexander-schranz , updated ;-)

cirykpopeye: > @cirykpopeye Great thank you this fixes #244

I had to add a \ in front of stdClass, I updated it in a push below.

PR #251, opened 2020-05-05
Fix dynamic list with media

### LINKED ISSUE #249: DynamicListBuilder breaks Listview
| Q | A
| --- | ---
| Bug? | Yes
| New Feature? | no
| Bundle Version | 9a43be625bbdf32e9a776782ec85fe8995934cb5 but c7f01fd47656bdfda4c82d6d25bb62db067c5f48 also has this issue
| Sulu Version | 2.0.6

#### Actual Behavior

When adding a Attachment Field and uploading this by a frontend form, the List View will Error with the following message:
```
Argument 1 passed to Sulu\Bundle\FormBundle\ListBuilder\DynamicListBuilder::getMediaUrls() must be of the type string, array given, called in /.../vendor/sulu/form-bundle/ListBuilder/DynamicListBuilder.php on line 57
```

In the Symfony Profiler nothing seems to be wrong during file upload.

```
[▼
  "attachment" => [▼
    Symfony\Component\HttpFoundation\File\UploadedFile {#25 ▼
      -test: false
      -originalName: "jonatan-pie-C2YORgsJqXs-unsplash (1).jpg"
      -mimeType: "image/jpeg"
      -error: 0
      path: "/private/var/tmp"
      filename: "phpkV9Xb3"
      basename: "phpkV9Xb3"
      pathname: "/private/var/tmp/phpkV9Xb3"
      extension: ""
      realPath: "/private/var/tmp/phpkV9Xb3"
      aTime: 2020-05-05 14:59:29
      mTime: 2020-05-05 14:59:29
      cTime: 2020-05-05 14:59:29
      inode: 42311052
      size: 1584286
      perms: 0100600
      owner: 1924723254
      group: 0
      type: "file"
      writable: true
      readable: true
      executable: false
      file: true
      dir: false
      link: false
    }
  ]
]
```

In the database the file looks as follows: `attachment: [20]`

#### Expected Behavior

List View should display normally.

#### Steps to Reproduce

As Stated above create a new Dynamic form with an attachment field in it. Display this form somewhere in the Frontend. Then upload a file using the field (we used Dropzone as plugin, i did not test html5 file input). After this the list view should break and throw a 500.

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #249 
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

Use a number for the url generation.

#### Why?

Fix dynamic list with media

### DIFF
--- a/ListBuilder/DynamicListBuilder.php
+++ b/ListBuilder/DynamicListBuilder.php
@@ -23,17 +23,17 @@ class DynamicListBuilder implements DynamicListBuilderInterface
     /**
      * @var string
      */
-    protected $delimiter;
+    private $delimiter;
 
     /**
      * @var RouterInterface
      */
-    protected $router;
+    private $router;
 
     /**
      * @var string
      */
-    protected $downloadUrl;
+    private $downloadUrl;
 
     public function __construct(string $delimiter, RouterInterface $router)
     {
@@ -70,7 +70,7 @@ class DynamicListBuilder implements DynamicListBuilderInterface
      *
      * @param mixed $value
      */
-    protected function toString($value): string
+    private function toString($value): string
     {
         if (is_string($value) || is_numeric($value)) {
             return $value;
@@ -100,7 +100,7 @@ class DynamicListBuilder implements DynamicListBuilderInterface
      *
      * @return string
      */
-    protected function getMediaUrls($value): string
+    private function getMediaUrls($value): string
     {
         if (is_string($value)) {
             return $this->getMediaUrl($value);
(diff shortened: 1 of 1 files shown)


ListBuilder/DynamicListBuilder.php
```diff
@@ -117,25 +117,28 @@ protected function getMediaUrls($value): string
...
     /**
      * For performance generate route only once.
      */
-    protected function getDownloadUrl(): string
+    private function getDownloadUrl(): string
     {
         if (null === $this->downloadUrl) {
-            $this->downloadUrl = $this->router->generate(
+            // The given id must be a number which we replace
+            $idReplacerNumber = '875421';
```
wachterjohannes, line 132: dont get it - why is this a fixed number?

alexander-schranz: As the route generater is expensive process and should be avoided todo multiple times, its only done once and the number is replaced with `{id}`. In 1.6 it was possible to directly use `{id}` but as the route now requires a number it now used a dummy number to be replaced later with the `{id}` replacer.

wachterjohannes: OK 🙈

PR #255, opened 2020-05-07
Avoid magic setter and getter in dynamic form

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| License | MIT

#### What's in this PR?

Avoid magic setter and getter in dynamic form.

#### Why?

Magic setter and getter should always be avoided.

### DIFF
--- a/Entity/Form.php
+++ b/Entity/Form.php
@@ -203,7 +203,7 @@ class Form
             $value = null;
 
             if ($dynamic) {
-                $value = $dynamic->{$field->getKey()};
+                $value = $dynamic->getField($field->getKey());
             }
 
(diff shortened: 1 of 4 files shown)


Entity/Form.php
```diff
@@ -203,7 +203,7 @@ public function serializeForLocale(string $locale, ?Dynamic $dynamic = null): ar
             $value = null;
 
             if ($dynamic) {
-                $value = $dynamic->{$field->getKey()};
```
nnatter, line 206: 😮

PR #267, opened 2020-09-25
Add possibility for a honeypot field for 1.x

Quehnie:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes (copy from version for Sulu 2)
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Also add the honeypot field possibility for this older version.

#### Why?

Spam problems.

#### Example Usage

In documentation

### DIFF
--- a/Form/Handler.php
+++ b/Form/Handler.php
@@ -75,6 +75,8 @@ class Handler implements HandlerInterface
      * @param EngineInterface $templating
      * @param EventDispatcherInterface $eventDispatcher
      * @param MediaManager $mediaManager
+     * @param string $honeyPotStrategy
+     * @param string $honeyPotField
      */
     public function __construct(
         ObjectManager $entityManager,
@@ -82,8 +84,8 @@ class Handler implements HandlerInterface
         EngineInterface $templating,
         EventDispatcherInterface $eventDispatcher,
         MediaManager $mediaManager,
-        string $honeyPotStrategy = self::HONEY_POT_STRATEGY_SPAM,
-        string $honeyPotField = null
+        $honeyPotStrategy = self::HONEY_POT_STRATEGY_SPAM,
+        $honeyPotField = null
     ) {
         $this->entityManager = $entityManager;
         $this->mailHelper = $mailHelper;
@@ -168,7 +170,6 @@ class Handler implements HandlerInterface
     {
         if ($adminMailConfiguration = $configuration->getAdminMailConfiguration()) {
             $subjectPrefix = '';
-
             if ($this->isSpamByHoneypot($form)) {
                 $subjectPrefix = '(SPAM) ';
             }
@@ -188,7 +189,7 @@ class Handler implements HandlerInterface
      * @param MailConfigurationInterface $configuration
      * @param string $subjectPrefix
      */
-    private function sendMail(FormInterface $form, MailConfigurationInterface $configuration, string $subjectPrefix = '')
+    private function sendMail(FormInterface $form, MailConfigurationInterface $configuration, $subjectPrefix = '')
     {
         // TODO FIXME this is currently the only way to get the medias to the email view.
         $additionalData = [];
@@ -341,7 +342,7 @@ class Handler implements HandlerInterface
             )
         );
     }
-    
+
     private function isSpamByHoneypot(FormInterface $form): bool
     {
         if (!$this->honeyPotField) {
--- a/Form/Handler.php
+++ b/Form/Handler.php
@@ -60,6 +60,16 @@ class Handler implements HandlerInterface
     protected $attachments = [];
 
     /**
+     * @var string
+     */
+    private $honeyPotStrategy;
+
+    /**
+     * @var string|null
+     */
+    private $honeyPotField;
+
+    /**
      * @param ObjectManager $entityManager
      * @param Mail\HelperInterface $mailHelper
      * @param EngineInterface $templating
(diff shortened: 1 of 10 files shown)


Form/Type/DynamicFormType.php
```diff
@@ -149,6 +157,21 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+        if ($this->honeyPotField) {
+            $builder->add(
+                str_replace(' ', '_', strtolower($this->honeyPotField)),
+                EmailType::class,
+                [
+                    'label' => $this->honeyPotField,
+                    'mapped' => false,
+                    'attr' => [
+                        'honeypot' => true
+                    ],
```
alexander-schranz, line 169: attribute should not be here.

Instead it should be:

```php
'block_name' => 'honeypot',
```

See https://symfony.com/doc/3.4/reference/forms/types/form.html#block-name

The block_name should as in the docs be then `_honeypot_row`.

Resources/doc/dynamic.md
```diff
@@ -163,6 +163,63 @@ php bin/adminconsole debug:container --tag=sulu_form.dynamic.type
...
+```
+
+And then hide it in your css with:
+
+```css
+.honung {
+    display: none;
+}
+```
+
```
alexander-schranz, line 221: This should be the same as here: https://github.com/sulu/SuluFormBundle/blob/master/Resources/doc/dynamic.md#style-honey-pot-field

The block name should be something like `_honeypot_row`

alexander-schranz (changes requested):

Form/Type/DynamicFormType.php
```diff
@@ -149,6 +157,19 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
         ]);
 
+        if ($this->honeyPotField) {
+            $builder->add(
+                str_replace(' ', '_', strtolower($this->honeyPotField)),
+                EmailType::class,
+                [
+                    'label' => $this->honeyPotField,
+                    'mapped' => false,
+                    'block_name' => 'honeypot',
```
alexander-schranz, line 167: `block _honeypot_row` didn't work? then this can be removed

Form/Handler.php
```diff
@@ -71,14 +81,18 @@ public function __construct(
         Mail\HelperInterface $mailHelper,
         EngineInterface $templating,
         EventDispatcherInterface $eventDispatcher,
-        MediaManager $mediaManager
+        MediaManager $mediaManager,
+        string $honeyPotStrategy = self::HONEY_POT_STRATEGY_SPAM,
+        string $honeyPotField = null
```
alexander-schranz, line 86: PHP Docs missing:

```
     * @param string $honeyPotStrategy
     * @param string $honeyPotField
```

Form/Handler.php
```diff
@@ -71,14 +81,18 @@ public function __construct(
         Mail\HelperInterface $mailHelper,
         EngineInterface $templating,
         EventDispatcherInterface $eventDispatcher,
-        MediaManager $mediaManager
+        MediaManager $mediaManager,
+        string $honeyPotStrategy = self::HONEY_POT_STRATEGY_SPAM,
```
alexander-schranz, line 85: `string` can not be used in all supported php versions and need to be removed

Form/Handler.php
```diff
@@ -155,8 +186,9 @@ private function sendMails(FormInterface $form, FormConfigurationInterface $conf
      *
      * @param FormInterface $form
      * @param MailConfigurationInterface $configuration
+     * @param string $subjectPrefix
      */
-    private function sendMail(FormInterface $form, MailConfigurationInterface $configuration)
+    private function sendMail(FormInterface $form, MailConfigurationInterface $configuration, string $subjectPrefix = '')
```
alexander-schranz, line 191: `string` typehint can not be used in this PHP version and need to be removed.

Form/Handler.php
```diff
@@ -142,7 +167,13 @@ private function save(FormInterface $form, FormConfigurationInterface $configura
     private function sendMails(FormInterface $form, FormConfigurationInterface $configuration)
     {
         if ($adminMailConfiguration = $configuration->getAdminMailConfiguration()) {
-            $this->sendMail($form, $adminMailConfiguration);
+            $subjectPrefix = '';
+
```
alexander-schranz, line 171: remove this empty (avoids us conflicts when merging into 2.x branch)

Form/Handler.php
```diff
@@ -303,4 +341,25 @@ protected function getPlainText(FormInterface $form, MailConfigurationInterface
             )
         );
     }
+
```
alexander-schranz, line 344: in the merge it looks like this line contains spaces remove the spaces if that is the case.

Resources/doc/dynamic.md
```diff
@@ -163,6 +163,52 @@ php bin/adminconsole debug:container --tag=sulu_form.dynamic.type
...
+
```
alexander-schranz, line 211: Only one empty line should be here.

alexander-schranz (changes requested):
Some small changes.

Form/Type/DynamicFormType.php
Quehnie: it does work but with prefix of the form so _dynamic_honeypot_row

alexander-schranz: okay

Form/Type/DynamicFormType.php
```diff
@@ -42,10 +48,12 @@ class DynamicFormType extends AbstractType
      */
     public function __construct(
         FormFieldTypePool $typePool,
-        Checksum $checksum
+        Checksum $checksum,
+        ?string $honeyPotField = null
```
alexander-schranz, line 52: This typehint should also be removed and `string|null $honeyPotField` added to the construct function

alexander-schranz (changes requested):

PR #276, opened 2021-01-08
Expand composer lint script and add php-cs-fixer as dev-dependency

niklasnatter:


composer.json
```diff
@@ -74,40 +75,43 @@
...
         "lint": [
-            "@lint-composer",
-            "@lint-yaml",
-            "@lint-twig",
-            "@lint-schema",
             "@phpstan",
-            "@lint-container"
-        ],
-        "lint-composer": "@composer validate --strict",
-        "lint-yaml": "@php Tests/Application/bin/adminconsole lint:yaml Resources/config",
```
niklasnatter, line 86: removed this because the `Resources/config` folder does not contain any yaml files

composer.json
```diff
@@ -74,40 +75,43 @@
         }
     },
     "scripts": {
+        "bootstrap-test-environment": [
+            "Tests/Application/bin/adminconsole doctrine:database:create --if-not-exists --env test",
+            "Tests/Application/bin/adminconsole doctrine:schema:update --force --env test"
+        ],
         "lint": [
```
niklasnatter, line 82: i changed the order of the scripts to match the other projects 🙂

composer.json
alexander-schranz: okay for me

alexander-schranz (approved):

PR #285, opened 2021-04-12
Fix csrf token validation after submit

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

Fix csrf token validation after submit.

#### Why?

Currently when submitted it does not complete correctly and errors with should not be called method.

### DIFF
--- a/Resources/config/services.xml
+++ b/Resources/config/services.xml
@@ -90,6 +90,7 @@
             <argument type="service" id="sulu_form.repository.form" />
             <argument type="service" id="form.factory" />
             <argument type="service" id="sulu_form.checksum" />
+            <argument type="service" id="security.csrf.token_manager"/>
         </service>
 
         <!-- Dynamic Form Type -->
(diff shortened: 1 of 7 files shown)


Resources/config/services.xml
```diff
@@ -90,6 +90,7 @@
             <argument type="service" id="sulu_form.repository.form" />
             <argument type="service" id="form.factory" />
             <argument type="service" id="sulu_form.checksum" />
+            <argument type="service" id="security.csrf.token_manager"/>
```
Prokyonn, line 93: Shouldn't we document this BC in the `UPGRADE.md` ?

Prokyonn (approved):

Resources/config/services.xml
alexander-schranz: 👍 added

Issue #44, opened 2017-02-01
Success redirect by internal link


alexander-schranz:
Instead of redirecting to a `?send=true` it should be possible for the content manager to choose a internal link to redirect to.

nvdbeek:
Is this feature coming in future releases? 

What is your idea by redirecting by internal link? By posting form data that you show a succes notice and then redirect or posting form data and redirect to the 'success' page where the notice is showing?

wachterjohannes:
the idea was to replace the `target` by a field that is able to select pages, articles or an external link. this is planned but was never requested by others :)

nvdbeek:
We would appreciate if this feature could be included.

alexander-schranz:
My suggestion to this feature as written in #287 would be possible to define the target page on the page not the form level, a form can be used in multiple webspaces where very page / webspace could have its own target / succes page:


```xml
        <property name="form" type="single_form_selection">
            <meta>
                <title lang="de">Formular</title>
                <title lang="en">Form</title>
            </meta>

            <params>
                <param name="resourceKey" value="page"/>
            </params>
        </property>

        <property name="targetPage" type="link">
            <meta>
                <title lang="de">Zielseite</title>
                <title lang="en">Targetpage</title>
            </meta>
            
            <tag name="sulu_form.target" form="form" />
        </property>
```

PR #287, opened 2021-05-04
Add ability to define landing page when form is submitted successfully

eekes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | #44
| License | MIT

#### What's in this PR?

Add the ability to define a custom landing page after a form was submitted successfully.

This is my first PR for a feature like this, so I'm not sure if I'm on the right track or whether this should be handled completely differently, so feel free to tear this down if needed... 😄 

#### To Do

- [ ] Create documentation
- [ ] Also provide a field to add a custom querystring to the landing page URL
- [ ] Tests (?)

### DIFF
--- a/Configuration/FormConfiguration.php
+++ b/Configuration/FormConfiguration.php
@@ -37,6 +37,11 @@ class FormConfiguration implements FormConfigurationInterface
     private $fileSave = true;
 
     /**
+     * @var string|null
+     */
+    private ?string $successRedirect = null;
+
+    /**
      * @var MailConfigurationInterface
      */
     private $adminMailConfiguration;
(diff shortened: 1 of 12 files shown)


alexander-schranz:
@eekes Thank you for the PR. I'm not really sure if we want to implment the target on the `Form` Entity, I did think in the past about it and I think it better would be that it can be defined per page as form can be used on different pages and could have different target pages. But we would need to discuss this first in the core team before we can create a decision what would be best to implement this feature.

eekes:
@alexander-schranz I hadn't thought about that but you're right, it would be better to be able to define this from wherever you select the form...

alexander-schranz:
Could think about something like this by using a `tag` currently not sure if possible to implement it this way:

```xml
        <property name="form" type="single_form_selection">
            <meta>
                <title lang="de">Formular</title>
                <title lang="en">Form</title>
            </meta>

            <params>
                <param name="resourceKey" value="page"/>
            </params>
        </property>

        <property name="targetPage" type="link">
            <meta>
                <title lang="de">Zielseite</title>
                <title lang="en">Targetpage</title>
            </meta>
            
            <tag name="sulu_form.target" form="form" />
        </property>
```

Update: A new `link` content type was added which would possible to link to page, articles, external or other link providers.

PR #289, opened 2021-05-21
Add protected media collection setting for form system collections

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

Add protected form collection setting.

#### Why?

In some cases you want that the uploaded media is only download able from the admin context in all cases.

#### Example Usage

```yml
sulu_form:
    media:
        protected: true
```

1. Create form with media (e.g.: `bin/console sulu:form:generate-form`)
2. Submit a media
3. Try to download the media you should be redirect to the admin url which will download the media only when you are logged in

#### Todo

 - [x] Create 2.2 Form Bundle registry recipe with this option enabled

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "doctrine/collections": "^1.0",
         "doctrine/orm": "^2.5.3",
         "doctrine/persistence": "^1.3 || ^2.0",
-        "sulu/sulu": "^2.0.3 || 2.3@dev",
+        "sulu/sulu": "^2.2.0 || 2.3@dev",
         "symfony/config": "^4.3 || ^5.0",
         "symfony/console": "^4.3 || ^5.0",
         "symfony/dependency-injection": "^4.3 || ^5.0",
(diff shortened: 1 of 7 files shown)


composer.json
```diff
@@ -21,7 +21,7 @@
         "doctrine/collections": "^1.0",
         "doctrine/orm": "^2.5.3",
         "doctrine/persistence": "^1.3 || ^2.0",
-        "sulu/sulu": "^2.0.3 || 2.3@dev",
+        "sulu/sulu": "^2.2.0 || 2.3@dev",
```
alexander-schranz, line 24: required for the media `$adminUrl` which was in 2.2.0 included

alexander-schranz:
Waiting here for some feedback from our partner @massiveart.

wachterjohannes (approved):

PR #291, opened 2021-06-09
EventListener only for download routes

Quehnie:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fixed that the event-listener is just executed for media routes.

#### Why?

It also reacted on the get_contact route for example which caused errors.


alexander-schranz (approved):

alexander-schranz:
@Quehnie Thank you!

PR #297, opened 2021-08-25
Improve documentation about TestFrom creation

henri9813:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Update the documentation about the TestForm creation

#### Why?

Currently, documentation tell that to CREATE A FORM, we should run this command, this is totally wrong, it will just create / update a form called `TestForm`. Which is not the same thing.

With the current documentation, we expect to have such a "interactive builder" of a form like the make:entity from SYmfony


niklasnatter:
Sounds reasonable to me - thanks for changing this! 🙂

PR #303, opened 2021-10-22
Add example and default value to radio buttons

Quehnie:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | yes
| License | MIT

#### What's in this PR?

Content managers can add a default value and a placeholder for radiobuttons.

#### Why?

If nothing is set there is a "None" value shown as default.

#### Example Usage

Create a new form with radio buttons.

#### BC Breaks/Deprecations

~~Existing "None" selections will dissapear.~~ EDIT: this is expected and can be avoided by the content manager when selecting `Required` on the radio buttons, or any other choice.

### DIFF
--- a/Dynamic/Types/ChoiceTrait.php
+++ b/Dynamic/Types/ChoiceTrait.php
@@ -39,8 +39,6 @@ trait ChoiceTrait
         FormFieldTranslation $translation,
         array $options
     ): array {
-        $options['placeholder'] = false;
-
         if (isset($options['attr']['placeholder'])) {
             $options['placeholder'] = $options['attr']['placeholder'];
             unset($options['attr']['placeholder']);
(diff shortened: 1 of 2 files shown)


Dynamic/Types/ChoiceTrait.php
```diff
@@ -39,6 +39,8 @@ private function getChoiceOptions(
         FormFieldTranslation $translation,
         array $options
     ): array {
+        $options['placeholder'] = false;
```
alexander-schranz: if no empty value should be displayed the content manager need to set the value the `required` flag. Will remove this change.

alexander-schranz (approved):

alexander-schranz:
@Quehnie Thank you!

PR #305, opened 2021-11-16
Fix typehint of Dynamic::getAttachment()

luca-rath:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

Fix typehint of `Dynamic::getAttachment()`

### DIFF
--- a/Entity/Dynamic.php
+++ b/Entity/Dynamic.php
@@ -361,9 +361,9 @@ class Dynamic implements AuditableInterface
     }
 
     /**
-     * @return string
+     * @return string|mixed[]|null
      */
-    public function getAttachment(): ?string
+    public function getAttachment()
     {
         return $this->getField('attachment');
     }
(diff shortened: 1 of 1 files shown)


Entity/Dynamic.php
```diff
@@ -361,9 +361,9 @@ public function getDate(): ?string
     }
 
     /**
-     * @return string
+     * @return string|mixed[]|null
```
luca-rath: @alexander-schranz I'm not sure what this method will return in all states of a request ... I know in a event listener for `FormSavePostEvent`, this will return `int[]`

alexander-schranz: I think this is always `null|int[]` and attachment is never a string, as the method is not used anywhere in the core this did not yet throw an error.

luca-rath: fixed

alexander-schranz (approved):

PR #306, opened 2021-11-16
Allow adding errors in FormSavePreEvent

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

This pr allows to add errors in `FormSavePreEvent` listener

### DIFF
--- a/Form/Handler.php
+++ b/Form/Handler.php
@@ -103,7 +103,9 @@ class Handler implements HandlerInterface
             $this->mapMediaIds($form->getData(), $mediaIds);
         }
 
-        $this->save($form, $configuration);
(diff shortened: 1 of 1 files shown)


alexander-schranz:
As discussed this should be handled over the [Symfony Form PreSubmit event](https://symfony.com/doc/5.3/form/events.html#a-the-formevents-pre-submit-event) as we want to avoid that invalid form creates Media Entities, which would be possible if we allow adding error at this stage.

PR #307, opened 2021-11-30
Fix Symfony 5.4 compatibility of DisabledCsrfTokenManager

eekes:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Make DisabledCsrfTokenManager compatible with Symfony's CsrfTokenManagerInterface

#### Why?

DisabledCsrfTokenManager doesn't seem be compatible with Symfony 5.4 because of the return types that were added.

```
Compile Error: Declaration of Sulu\Bundle\FormBundle\Csrf\DisabledCsrfTokenManager::getToken($tokenId) must be compatible with Symfony\Component\Security\Csrf\CsrfTokenManagerInterface::getToken(string $tokenId): Symfony\Component\Security\Csrf\CsrfToken
```

#### Example Usage

Simply by using a form in a Symfony 5.4 Sulu project.

#### BC Breaks/Deprecations

Don't think so?


alexander-schranz:
Looks for me correctly for [5.4](https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Security/Csrf/CsrfTokenManagerInterface.php) but it seems to install a newer version. So I think our error is that we forget to define the dependency to csrf token package.

alexander-schranz:
Can you add `"symfony/security-csrf": "^4.4 || ^5.4"` to the `composer.json` then there should be no adjustments needed. We got [problems](https://github.com/sulu/SuluFormBundle/pull/292) when adding the return types on PHP 7.2 on Symfony 4.4 so we can not add them there.

alexander-schranz (approved):

eekes:
That seems to fix it, installed 6.0 without defining the version in composer.

alexander-schranz (approved):

alexander-schranz:
@eekes Thx for the report and the pull request.

alexander-schranz:
@eekes released as 2.2.1.

PR #310, opened 2022-01-20
Add sendinblue integration

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Add sendinblue integration

#### Example Usage

See `sendinblue.md` file

#### To Do

- [x] Create documentation


Resources/config/form-fields/field_sendinblue.xml
```diff
@@ -0,0 +1,46 @@
...
+    <property name="options/attributeNameFirstName" type="text_line" colspan="6">
+        <meta>
+            <title>sulu_form.sendinblue_attribute_name_first_name</title>
+        </meta>
+    </property>
+    <property name="options/attributeNameLastName" type="text_line" colspan="6">
+        <meta>
+            <title>sulu_form.sendinblue_attribute_name_last_name</title>
+        </meta>
+    </property>
```
alexander-schranz: This should be as simple as possible and we should here just go in the listener with FIRST_NAME and LAST_NAME defaults like we do it in mailchimp also.

Resources/config/form-fields/field_sendinblue.xml
```diff
@@ -0,0 +1,46 @@
...
+                    type="expression"
+                    value="service('sulu_form.dynamic.sendinblue_mail_template_select').getValues()"
+            />
+        </params>
+    </property>
+    <property name="options/redirectionUrl" type="url" mandatory="true">
+        <meta>
+            <title>sulu_form.sendinblue_redirection_url</title>
+        </meta>
+    </property>
```
alexander-schranz: What is this for? Do we need it? Can we just not use the current page url or even no url. Should not sending blue UI define this.

luca-rath: This is required by the sendinblue api ... But yes, I could just use the current url

luca-rath:
@alexander-schranz fixed everything

Dynamic/Helper/SendinblueListSelect.php
```diff
@@ -0,0 +1,51 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\FormBundle\Dynamic\Helper;
+
+use SendinBlue\Client\Api\ContactsApi;
+use SendinBlue\Client\Configuration;
+
+class SendinblueListSelect
```
alexander-schranz:
```suggestion

/**
 * @final
 * @internal
 */
class SendinblueListSelect
```

Dynamic/Helper/SendinblueMailTemplateSelect.php
```diff
@@ -0,0 +1,55 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\FormBundle\Dynamic\Helper;
+
+use SendinBlue\Client\Api\TransactionalEmailsApi;
+use SendinBlue\Client\Configuration;
+
+class SendinblueMailTemplateSelect
```
alexander-schranz:
```suggestion

/**
 * @final
 * @internal
 */
class SendinblueMailTemplateSelect
```

Event/SendinblueListSubscriber.php
```diff
@@ -0,0 +1,110 @@
...
+namespace Sulu\Bundle\FormBundle\Event;
+
+use SendinBlue\Client\Api\ContactsApi;
+use SendinBlue\Client\Configuration;
+use SendinBlue\Client\Model\CreateDoiContact;
+use Sulu\Bundle\FormBundle\Entity\Dynamic;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpFoundation\RequestStack;
+
+class SendinblueListSubscriber implements EventSubscriberInterface
```
alexander-schranz:
```suggestion

/**
 * @final
 * @internal
 */
class SendinblueListSubscriber implements EventSubscriberInterface
```

Event/SendinblueListSubscriber.php
```diff
@@ -0,0 +1,110 @@
...
+        if (!$request) {
+            return;
+        }
+
+        $form = $dynamic->getForm()->serializeForLocale($dynamic->getLocale(), $dynamic);
+
+        $email = '';
+        $firstName = '';
+        $lastName = '';
+        $redirectionUrl = $request->getUriForPath('');
```
alexander-schranz: @luca-rath should we add`?subscribe=true`. So in the frontend a message can be shown?

alexander-schranz (changes requested):
Some small change requests.

Event/SendinblueListSubscriber.php
luca-rath:
```suggestion
        $redirectionUrl = $request->getUriForPath('') . '?subscribe=true';
```

alexander-schranz (approved):

PR #315, opened 2022-02-02
Allow multiple form data tabs on parent view

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

Allow multiple form data tabs on parent view

### DIFF
--- a/Admin/DynamicListAdmin.php
+++ b/Admin/DynamicListAdmin.php
@@ -46,6 +46,7 @@ class DynamicListAdmin extends Admin
     public function configureViews(ViewCollection $viewCollection): void
     {
         foreach ($this->config as $parent => $sections) {
+            /** @var mixed[] $config */
             foreach ($sections as $config) {
                 if (!$viewCollection->has($parent)) {
                     continue;
--- a/Admin/DynamicListAdmin.php
+++ b/Admin/DynamicListAdmin.php
@@ -46,18 +46,20 @@ class DynamicListAdmin extends Admin
     public function configureViews(ViewCollection $viewCollection): void
     {
         foreach ($this->config as $parent => $sections) {
+            $counter = 0;
             /** @var mixed[] $config */
-            foreach ($sections as $config) {
+            foreach ($sections as $key => $config) {
+                 ++$counter;
                 if (!$viewCollection->has($parent)) {
                     continue;
                 }
 
                 $action = 'form-list';
-                if (isset($config['action'])) {
-                    $action = $config['action'];
+                if ($counter > 1) {
+                    $action .= '-' . str_replace('_', '-', $key);
                 }
 
-                $name = $parent . '.' . $action;
+                $name = $parent . '.' . $action . '-key';
                 if (isset($config['name'])) {
                     $name = $config['name'];
                 }
--- a/Admin/DynamicListAdmin.php
+++ b/Admin/DynamicListAdmin.php
@@ -45,14 +45,17 @@ class DynamicListAdmin extends Admin
 
     public function configureViews(ViewCollection $viewCollection): void
     {
+        /** @var array<string, mixed> $sections */
         foreach ($this->config as $parent => $sections) {
+            if (!$viewCollection->has($parent)) {
+                continue;
+            }
+
             $counter = 0;
+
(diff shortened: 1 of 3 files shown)


alexander-schranz (changes requested):
@luca-rath the new poroperty should also be added to the documentation: https://github.com/sulu/SuluFormBundle/blob/2.x/Resources/doc/dynamic.md#list-tab---export

Resources/doc/dynamic.md
```diff
@@ -132,6 +132,7 @@ sulu_form:
     dynamic_lists:
         sulu_page.page_edit_form: # view key e.g. "sulu_page.page_edit_form" for sulu pages or "sulu_article.edit_form" for sulu articles.
             form: # unique key mostly the same as the template key or a combination between template and property key.
```
alexander-schranz: as reading this command I think we should just use the `key` as postfix for the viewbuilder, so uniqueness is defined by the yaml key.

Admin/DynamicListAdmin.php
```diff
@@ -45,20 +45,27 @@ public function __construct(ViewBuilderFactoryInterface $viewBuilderFactory, arr
 
     public function configureViews(ViewCollection $viewCollection): void
     {
-        $action = 'form-list';
-        // TODO handle multipage views
-
         foreach ($this->config as $parent => $sections) {
-            foreach ($sections as $key => $config) {
+            /** @var mixed[] $config */
+            foreach ($sections as $config) {
```
alexander-schranz:
```suggestion
            foreach ($sections as $key => $config) {
                 ++$counter;
```

Admin/DynamicListAdmin.php
```diff
@@ -45,20 +45,27 @@ public function __construct(ViewBuilderFactoryInterface $viewBuilderFactory, arr
...
                 if (!$viewCollection->has($parent)) {
                     continue;
                 }
 
-                $name = $parent . '.' . $action . '-key';
+                $action = 'form-list';
+                if (isset($config['action'])) {
+                    $action = $config['action'];
+                }
```
alexander-schranz: Following way we avoid the bc break when using the key:

```suggestion
                $action = 'form-list';
                if ($counter > 1) {
                    $action .= '-' . str_replace('_', '-', $key);
                }
```

Admin/DynamicListAdmin.php
```diff
@@ -45,20 +45,27 @@ public function __construct(ViewBuilderFactoryInterface $viewBuilderFactory, arr
 
     public function configureViews(ViewCollection $viewCollection): void
     {
-        $action = 'form-list';
-        // TODO handle multipage views
-
         foreach ($this->config as $parent => $sections) {
```
alexander-schranz, line 49:
```suggestion
        foreach ($this->config as $parent => $sections) {
            $counter = 0;
```

Admin/DynamicListAdmin.php
```diff
@@ -45,20 +45,27 @@ public function __construct(ViewBuilderFactoryInterface $viewBuilderFactory, arr
...
                     continue;
                 }
 
-                $name = $parent . '.' . $action . '-key';
+                $action = 'form-list';
+                if (isset($config['action'])) {
+                    $action = $config['action'];
+                }
+
+                $name = $parent . '.' . $action;
```
alexander-schranz: Required to avoid a bc break of the view name:

```suggestion
                $name = $parent . '.' . $action . '-key';
```

alexander-schranz (changes requested):
After rethinking I think we should use the `key` instead for action and view name as that make sure that is unique by the yaml structure. Sorry for the circumstances!

I added some suggestion so we can still avoid bc break in url and view name.

Resources/doc/dynamic.md
```diff
@@ -132,6 +132,7 @@ sulu_form:
...
+                action: form-list # the path of the view
```
luca-rath:
```suggestion
```

alexander-schranz (approved):

PR #316, opened 2022-02-02
Add toolbar action for creating copy of current form to edit view

### LINKED ISSUE #151: Copy function of existing forms
| Q | A
| --- | ---
| Bug? | no
| New Feature? | yes
| Bundle Version | Specific version or SHA of a commit
| Sulu Version | Specific version or SHA of a commit
| Browser Version | Browser name and version

#### Actual Behavior

How does it behave at the moment? 
At the moment, you have to rebuild every list, every time.

#### Expected Behavior

What is the behavior you expect?
It will save the user a lot of time!

#### Steps to Reproduce

image

#### Possible Solutions

If you have already ideas how to solve the issue, add them here.
(remove this section if not needed)

integrate the copy function in the header (right beside "add new") - similar to the copy function in the Content-Section

Danke Alex :-)

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #151
| Related issues/PRs | depends on https://github.com/sulu/sulu/pull/6475
| License | MIT

#### What's in this PR?

This pull requests adds a toolbar action for creating a copy of the current form to the form edit view. It depends on the toolbar action implemented in https://github.com/sulu/sulu/pull/6475.

#### Why?

See #151

### DIFF
--- a/Controller/FormController.php
+++ b/Controller/FormController.php
@@ -179,34 +179,11 @@ class FormController extends AbstractRestController implements ClassResourceInte
     {
         $action = $request->query->get('action');
         $locale = $this->getLocale($request);
-        $entity = $this->formManager->findById($id, $locale);
-
-        if (!$entity) {
-            throw new NotFoundHttpException(sprintf('No form with id "%s" was found!', $id));
-        }
 
         try {
             switch ($action) {
                 case 'copy':
-                    $copiedForm = null;
-                    foreach ($entity->getTranslations() as $translation) {
-                        /** @var array{ title: string, fields: mixed[] } $data */
-                        $data = $this->getApiEntity($entity, $translation->getLocale());
-                        $data['title'] = $data['title'] . ' (2)';
-
-                        /** @var array{ options: \stdClass|mixed[] } $field */
-                        foreach ($data['fields'] as &$field) {
-                            if ($field['options'] instanceof \stdClass) {
-                                $field['options'] = [];
-                            }
-                        }
-
-                        $copiedForm = $this->formManager->save(
-                            $data,
-                            $translation->getLocale(),
-                            $copiedForm ? $copiedForm->getId() : null
-                        );
-                    }
+                    $copiedForm = $this->formManager->copy($id);
 
                     return $this->handleView($this->view($this->getApiEntity($copiedForm, $locale)));
                 default:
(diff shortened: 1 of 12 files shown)


Controller/FormController.php
```diff
@@ -174,6 +175,50 @@ public function postAction(Request $request): Response
...
+            switch ($action) {
+                case 'copy':
+                    $copiedForm = null;
+                    foreach ($entity->getTranslations() as $translation) {
+                        /** @var array{ title: string, fields: mixed[] } $data */
+                        $data = $this->getApiEntity($entity, $translation->getLocale());
+                        $data['title'] = $data['title'] . ' (2)';
+
+                        /** @var array{ options: \stdClass|mixed[] } $field */
+                        foreach ($data['fields'] as &$field) {
```
niklasnatter: this is needed because the `getApiEntity` method returns an empty `\stdClass` here: https://github.com/sulu/SuluFormBundle/blob/c3908bd532df7563759a6fcc270f762e9e1d40f9/Controller/FormController.php#L367

Controller/FormController.php
```diff
@@ -174,6 +175,50 @@ public function postAction(Request $request): Response
...
+
+        if (!$entity) {
+            throw new NotFoundHttpException(sprintf('No form with id "%s" was found!', $id));
+        }
+
+        try {
+            switch ($action) {
+                case 'copy':
+                    $copiedForm = null;
+                    foreach ($entity->getTranslations() as $translation) {
```
niklasnatter: i implemented this here because the existing `getApiEntity` method made it really easy to do it 😇 

should we create a `copy` method on the `FormManager` instead? if yes, we need to move the `getApiEntity` method to the `FormManager` too or we cannot easily reuse the existing save method

niklasnatter: this solution forces us to also dispatch the `FormCopiedEvent` inside of the `FormController`: 
https://github.com/sulu/SuluFormBundle/pull/318/files#diff-12e096b425e19f6ccc009af6a1dae8d0ecb19260daf4d3fa288558b41c6c7e14R226

this does not feel great and i guess it is a hint that it would be nicer to do this in the `FormManager`. what do you think?

alexander-schranz: I moved it to the manager. also it did make problems when using save when the form has multiple languages.

alexander-schranz (approved):

PR #326, opened 2022-04-21
Add trash functionality for forms

niklasnatter:
This PR integrates the form entity with the `SuluTrashBundle`

### DIFF
--- a/Entity/Form.php
+++ b/Entity/Form.php
@@ -100,6 +100,7 @@ class Form
             $translation = new FormTranslation();
             $translation->setLocale($locale);
             $this->addTranslation($translation);
+            $translation->setForm($this);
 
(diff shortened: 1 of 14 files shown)


Entity/Form.php
```diff
@@ -100,6 +100,7 @@ public function getTranslation(string $locale, bool $create = false, bool $fallb
             $translation = new FormTranslation();
             $translation->setLocale($locale);
             $this->addTranslation($translation);
+            $translation->setForm($this);
```
niklasnatter, line 103: the same thing is already done in the `FormField`:
https://github.com/sulu/SuluFormBundle/blob/1ab810f829233c6c006c2bac74406aead7a68137/Entity/FormField.php#L167

PR #328, opened 2022-07-14
Fix file type attr

RealAestan:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

The attribute max is removed on input type file
The attribute accept is removed when there is not restrictions on input file type

#### Why?

max attr is no only applicable on `date` `month` `week` `time` `datetime-local` `number` `range` input types
accept attr should not be empty, when all files are accepted the attribute should be removed


Dynamic/Types/AttachmentType.php
```diff
@@ -78,8 +80,6 @@ public function build(FormBuilderInterface $builder, FormField $field, string $l
             $options['constraints'][] = new Count([
                 'max' => $fileMax,
             ]);
-
-            $options['attr']['max'] = $fileMax;
```
alexander-schranz: Remove this would be a bc break and would crash form themes try to read this attribute to be displayed as value on the website.

At current state we can only add a comment to `// @deprecated` this and add a line to the `UPGRADE.md` to remove it in the next major.

For your website you can filter out that attribute in your custom form theme if you don't want to render it.

Dynamic/Types/AttachmentType.php
```diff
@@ -54,7 +54,9 @@ public function build(FormBuilderInterface $builder, FormField $field, string $l
             }
         }
 
-        $options['attr']['accept'] = \implode(',', $mimeTypes);
+        if (!empty($mimeTypes)) {
+            $options['attr']['accept'] = \implode(',', $mimeTypes);
+        }
```
alexander-schranz, line 59: 👍

alexander-schranz (changes requested):
Thx for the pull request. Little change request.

Dynamic/Types/AttachmentType.php
RealAestan: Hello, my bad I totally forgot that use case. I added a new "data-max" attribute to allow people who have this use case to have this possibility. Should I add something to UPGRADE.md or let you handle it when you will prepare the release ?

alexander-schranz: yeah a sentence in the UPGRADE.md would be great about it.

Dynamic/Types/AttachmentType.php
```diff
@@ -79,7 +81,9 @@ public function build(FormBuilderInterface $builder, FormField $field, string $l
                 'max' => $fileMax,
             ]);
 
+            // TODO : Remove when a new major is release and tell user to migrate to data-max attribute
```
alexander-schranz:
```suggestion
            // @deprecated remove when a new major is release as max is not supported by input[type=file]
```

alexander-schranz:
@RealAestan Thank you!

PR #329, opened 2022-07-14
Add compatibility for Symfony 6

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Add compatibility for Symfony 6.

#### Why?

Compatiblity to Symfony 6 and get ride of deprecated swiftmailer.


luca-rath (approved):

Tests/Application/config/swiftmailer.yml
```diff
@@ -0,0 +1,2 @@
+swiftmailer:
+    disable_delivery:  true
```
luca-rath, line 2: One question, why do we have a swiftmailer config, if swiftmailer isn't installed?

alexander-schranz: its installed when using Symfony 5.4 or Sulu 2.4 (lower dependencies)

alexander-schranz: and there its still the default mailer to not break projects which maybe did create swiftmailer listeners.

alexander-schranz: handled here: https://github.com/sulu/SuluFormBundle/blob/ba31c3e69d477b5665e24ffb23d91bbd446e6edc/DependencyInjection/SuluFormExtension.php#L244-L258

rogamoore:
Would it be possible to release these changes? I noticed that currently our projects are unable to update to Sulu 2.5 because it fails with

```
!!    Compile Error: Declaration of Sulu\Bundle\FormBundle\Controller\FormWebsite
!!    Controller::getSubscribedServices() must be compatible with Sulu\Bundle\Web
!!    siteBundle\Controller\WebsiteController::getSubscribedServices(): array
```

From the changes here I guess that should be fixed already. :)

alexander-schranz:
@rogamoore we want to first test the demo with all upgraded bundles before doing a stable release of the bundles. If you already want to use you can install this via using `^2.5@dev`  in your composer.json for the sulu/form-bundle.

PR #342, opened 2022-12-20
Make salutation gender neutral

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Add `Other` option to salutation field and add `Please choose` placeholder instead of automatically select `Mr.`.

#### Why?

To make salutation gender neutral.


Dynamic/Types/SalutationType.php
```diff
@@ -48,8 +49,9 @@ public function build(FormBuilderInterface $builder, FormField $field, string $l
     protected function getChoices(): array
     {
         return [
-            'sulu_form.salutation_mr' => 'mr',
             'sulu_form.salutation_ms' => 'ms',
+            'sulu_form.salutation_mr' => 'mr',
+            'sulu_form.salutation_other' => 'other',
```
alexander-schranz: I think this is not the correct way of handling this. As a `Salutation` is never gender neutral. The salutation is used in `Letters` and something like `Hello Other Alexander Schranz,` is I think not the expected result.

The correct way for me would be here make sure that there is no default and no `Please choose`, something like `No salutation` or `No choice` / `Keine auswahl` should be the default value of the form, with an empty value.

alexander-schranz (changes requested):

Dynamic/Types/SalutationType.php
alexander-schranz: Did just now see this is the `FormBundle` this salutations represent `Salutation` of ContactBundle and so this need to be changed there first.

alexander-schranz (approved):

PR #352, opened 2023-02-24
dynamic forms: data detail view

Basster:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | no
| Related issues/PRs | no
| License | MIT

#### What's in this PR?

One of my client wants to have a detail view of dynamic form submissions, because he doesn't want to export everything, but check the submissions from the Sulu Backend. If some text-editor field values are too long, the Table component just hides the text.

#### Why?

As I'm still pretty new to Sulu, maybe there is another, easier solution, without touching the admin frontend. So this PR is **more of a PoC and open for discussion**.

I didn't find an actual "view" or "readonly form" mode in Sulu, so I tried this approach.

#### Example Usage

Read-Only View:  
Bildschirm­foto 2023-02-24 um 13 08 44

Preview Button in list:  
image

#### BC Breaks/Deprecations

It adds admin frontend code

#### To Do

- [ ] Discuss better solution
- [ ] Field mappings
- [ ] Attachments
- [ ] Create documentation

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,0 +1,27 @@
+[options]
+types_first=false
+esproposal.decorators=ignore
+esproposal.optional_chaining=enable
+module.name_mapper='.+\.s?css' -> 'empty/object'
+module.name_mapper='fos-jsrouting/router' -> 'empty/object'
+
+[untyped]
+.*/node_modules/react-leaflet/.*
+
+[ignore]
+.*/node_modules/bower-json/.*
+.*/node_modules/config-chain/.*
+.*/node_modules/findup/.*
+.*/node_modules/jss/.*
+.*/node_modules/npmconf/.*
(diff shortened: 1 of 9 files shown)


alexander-schranz:
@Basster Sorry for the late response here. We currently avoid custom JS inside our core bundles.

I think best if we tackle how we can make in sulu core ([sulu/sulu](https://github.com/sulu/sulu)) maybe a `view` which is just used for displaying data some kind of `widgets` view and put there create some widgets which we could reuse here. Maybe we need a new type of `xml` to tackle that.

Feel free to create an issue in the sulu/sulu repository to discuss how such widgets view / readonly views could look like.

PR #357, opened 2023-04-06
Added redirect link to sendinblue

martinlagler:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Adds redirect link to sendinblue.

### DIFF
--- a/Event/SendinblueListSubscriber.php
+++ b/Event/SendinblueListSubscriber.php
@@ -44,9 +44,9 @@ class SendinblueListSubscriber implements EventSubscriberInterface
 
     public function __construct(
         RequestStack $requestStack,
-        ?LinkProviderPoolInterface $linkProviderPool = null,
         ?string $apiKey,
-        ?ClientInterface $client = null
+        ?ClientInterface $client = null,
+        ?LinkProviderPoolInterface $linkProviderPool = null
     ) {
         $this->requestStack = $requestStack;
         $this->linkProviderPool = $linkProviderPool;
--- a/Event/SendinblueListSubscriber.php
+++ b/Event/SendinblueListSubscriber.php
@@ -16,6 +16,7 @@ use SendinBlue\Client\Api\ContactsApi;
 use SendinBlue\Client\Configuration;
 use SendinBlue\Client\Model\CreateDoiContact;
 use Sulu\Bundle\FormBundle\Entity\Dynamic;
+use Sulu\Bundle\MarkupBundle\Markup\Link\LinkProviderPoolInterface;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpFoundation\RequestStack;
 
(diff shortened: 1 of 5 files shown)


Event/SendinblueListSubscriber.php
```diff
@@ -36,12 +37,19 @@ class SendinblueListSubscriber implements EventSubscriberInterface
...
     private $contactsApi;
 
+    /**
+     * @var LinkProviderPoolInterface
+     */
+    private $linkProviderPool;
+
     public function __construct(
         RequestStack $requestStack,
+        LinkProviderPoolInterface $linkProviderPool,
```
Prokyonn: @alexander-schranz can we do the BC break here or should we add a BC layer?

alexander-schranz: as it is not much work a bc layer should be added here. includes that this need to be the the last parameter here and be ` = null` by default

Event/SendinblueListSubscriber.php
```diff
@@ -126,4 +140,35 @@ public function listSubscribe(FormSavePostEvent $event): void
             $this->contactsApi->createDoiContact($createDoiContact);
         }
     }
+
+    /**
+     * @param array<string, ?string> $redirectLink
```
alexander-schranz: define this more specific

Event/SendinblueListSubscriber.php
```diff
@@ -97,6 +106,11 @@ public function listSubscribe(FormSavePostEvent $event): void
                 $mailTemplateId = $field['options']['mailTemplateId'] ?? null;
                 /** @var int|null $listId */
                 $listId = $field['options']['listId'] ?? null;
+                $redirectLink = $field['options']['redirectLink'] ?? null;
+
+                if ($redirectLink && $this->linkProviderPool) {
```
alexander-schranz: Move check for `$this->linkProviderPool` that way phpstan now:

```suggestion
                if ($redirectLink) {
```

Event/SendinblueListSubscriber.php
```diff
@@ -126,4 +140,44 @@ public function listSubscribe(FormSavePostEvent $event): void
...
+    {
+        if (!$redirectLink['provider']) {
+            return null;
+        }
+
+        if ('external' === $redirectLink['provider']) {
+            return $redirectLink['href'];
+        }
+
+        $linkProvider = $this->linkProviderPool->getProvider($redirectLink['provider']);
```
alexander-schranz, line 170:
```suggestion
        if (!$this->linkProviderPool) {
            return null;
        }

        $linkProvider = $this->linkProviderPool->getProvider($redirectLink['provider']);
```

alexander-schranz (changes requested):

alexander-schranz (approved):

PR #360, opened 2023-04-24
Fix DynamicController dynamicListFactory build arguments order

dev-newvisibility:
| Q | A
| --- | ---
| Bug fix? | yes
| License | MIT

#### What's in this PR?

DynamicController is passing arguments to the build method of the dynamicListFactory property in the wrong order. The locale should be passed as the second argument, and the builder alias as the third.


alexander-schranz (approved):

alexander-schranz:
@dev-newvisibility Thank you!

PR #361, opened 2023-04-24
Adds an option to export CSV with labels as heading

dev-newvisibility:
| Q | A
| --- | ---
| New feature? | yes
| Related issues/PRs | #359
| License | MIT

#### What's in this PR?

When exporting forms to CSV, the first row is filled with the field keys. This may be the preferable option if the CSV is to be read by a machine. However, if the CSV is to be processed by a human, a label is more helpful to distinguish the columns.
This PR adds an option to the sulu_form configuration to choose between the two formats.

With the option disabled (same as before):
nok

With the option enabled:
ok

### DIFF
--- a/Controller/DynamicController.php
+++ b/Controller/DynamicController.php
@@ -64,13 +64,19 @@ class DynamicController implements ClassResourceInterface
      */
     private $viewHandler;
 
+    /**
+     * @var bool
+     */
+    private $prefersLabeledExports;
+
     public function __construct(
         DynamicRepository $dynamicRepository,
         DynamicListFactory $dynamicListFactory,
         MediaManagerInterface $mediaManager,
         EntityManager $entityManager,
         FormRepository $formRepository,
-        ViewHandler $viewHandler
+        ViewHandler $viewHandler,
+        bool $prefersLabeledExports
     ) {
         $this->dynamicRepository = $dynamicRepository;
         $this->dynamicListFactory = $dynamicListFactory;
@@ -78,6 +84,7 @@ class DynamicController implements ClassResourceInterface
         $this->entityManager = $entityManager;
         $this->formRepository = $formRepository;
         $this->viewHandler = $viewHandler;
+        $this->prefersLabeledExports = $prefersLabeledExports;
     }
 
     /**
@@ -85,11 +92,12 @@ class DynamicController implements ClassResourceInterface
      */
     public function cgetAction(Request $request): Response
     {
+        $locale = $this->getLocale($request);
         $filters = $this->getFilters($request);
+        $view = $this->getView($request);
         $page = $request->get('page', 1);
         $limit = $request->get('limit');
         $offset = (int) (($page - 1) * $limit);
-        $view = $request->get('view', 'default');
         $sortOrder = $request->get('sortOrder', 'asc');
         $sortBy = $request->get('sortBy', 'created');
 
@@ -100,7 +108,7 @@ class DynamicController implements ClassResourceInterface
             $offset
         );
 
-        $entries = $this->dynamicListFactory->build($entries, $view);
+        $entries = $this->dynamicListFactory->build($entries, $locale, $view);
 
         // avoid total request when entries < limit
         if (\count($entries) == $limit) {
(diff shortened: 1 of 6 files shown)


Controller/DynamicController.php
```diff
@@ -100,7 +108,7 @@ public function cgetAction(Request $request): Response
             $offset
         );
 
-        $entries = $this->dynamicListFactory->build($entries, $view);
+        $entries = $this->dynamicListFactory->build($entries, $locale, $view);
```
rogamoore, line 111: Interestingly, this is actually fixing a bug. The second argument for the `build` method in `DynamicListFactoryInterface` is the locale. But until now, the `view` was passed instead.

```php
public function build(array $dynamics, string $locale, string $builder = 'default'): array;
```

The only reason it was working is because the third argument has a fallback and in the existing `DynamicListBuilder` the `locale` (which would be `default`) is not being used.

ListBuilder/DynamicLabeledListBuilder.php
```diff
@@ -0,0 +1,60 @@
...
+        foreach ($entry[0] as $key => $value) {
+            $entryLabeled[$labels[$key] ?? $key] = $value;
+        }
+
+        return [$entryLabeled];
+    }
+
+    private function getLabels(Form $form, string $locale): array
+    {
+        if (!isset($formFieldsLabelCache[$form->getId()])) {
```
rogamoore: Should be `$this->formFieldsLabelCache` I guess :)

dev-newvisibility: Fixed, thank you!

ListBuilder/DynamicLabeledListBuilder.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        if (!isset($this->formFieldsLabelCache[$form->getId()])) {
+            $labels = [];
+
+            foreach ($form->getFields() as $field) {
+                $label = '';
+                $translation = $field->getTranslation($locale, false, true);
+
+                if ($translation) {
+                    $label = $translation->getShortTitle() ?: \strip_tags($translation->getTitle());
```
rogamoore, line 52: I was testing this and I found an edge case: The "spacer" field does have a translation, but both `getShortTitle` and `getTitle` return `null`- in this case `strip_tags` will fail because it expects a string.

IMHO when there is no translation or all titles are `null`, I would just skip the loop instead of setting an empty string. Then in the `build` method above it will fall back to the key, which is better than nothing.

rogamoore: I found two more special fields that need different handling: `freeText` and `headline`.
Both are just there to add additional text to the form, but are no real fields. Before it appeared in the export as "freeText", "freeText1" with no content in the column etc. Now it's using the title which can potentially be a really long text. So either it should be skipped (it has no content anyway) or the key should be used instead.

rogamoore: There are some other special fields that I cannot test like recaptcha etc. which probably should be checked as well how it behaves with this new feature.

rogamoore: Looks like there is already a constant with these special fields: https://github.com/sulu/SuluFormBundle/blob/2.5/Entity/Dynamic.php#L36

And the `getFields` method has a `hideHidden` flag, which does exactly what we want - skip these fields 👍

Controller/DynamicController.php
rogamoore: I just noticed https://github.com/sulu/SuluFormBundle/pull/360 😄

ListBuilder/DynamicLabeledListBuilder.php
dev-newvisibility: Hi, sorry for the delay. I've updated the PR with your suggestions.

alexander-schranz:
I'm not sure if a config option is the best here 🤔  

We should have a general discussion about the CSV Export and Labels as sure FormBundle is one place but there are not of other CSV Exports which currently only output keys.

What I'm not sure if a config options is the right way here or if we should get that generally into the UI as an option. The keys make sense in case of exports done for imports or external systems, but are in case of the form bundle not always user friendly (e.g. checkbox, checkbox_1, ...).

oprichnik:
No news here ? This would be a really useful update .

alexander-schranz:
Hello @oprichnik,

Nobody yet did have a look how to solve this in a general enough way so it works for all kind of CSV exports the same way in the Sulu core. I think the best option would be if there is maybe in the CSV Export an option if "keys" or "labels" should be used for the export. As "labels" always don't work some bundles and even projects have imports where the key is used. I think example the redirect bundle where has also import. As said this should be tackled in the Sulu core not for one specific bundle only.

PR #363, opened 2023-07-04
Supporting multiple dynamic forms on the same page

dev-newvisibility:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adds the URL parameter `form=<form-id>` when a form is submitted in addition to `send=true`

#### Why?

At the moment, if you add multiple dynamic forms to the same template, or if multiple forms are displayed on the same page, you can't know which one has been submitted due to the fact that the `send=true` URL parameter value is the same for all forms.

#### Example Usage

```twig
{% if app.request.get('send') == 'true' and app.request.get('form') == view.form.entity.formId %}
     {{ view.form.entity.successText|raw }}
{% endif %}
```


dev-newvisibility:
We've updated the PR to ensure it is a non-breaking change.

PR #368, opened 2023-09-12
Load fallback translation after form copy

Quehnie:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fixes an error which occurs on form copy when the user locale does not exist.

#### Why?

Forms could not be copied when the translation for the user was not existing.


alexander-schranz:
Looks good thx @Quehnie

PR #358, opened 2023-04-11
Add support of recaptcha v3

bodabodah:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | #331
| License | MIT

#### What's in this PR?

Add support for Recaptcha V3

PR #373, opened 2023-11-28
DRAFT: Extend with recaptcha v3

Quehnie:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

The correct way to use EWZ Recaptcha  v3 with sulu form bundle

#### Why?

EWZ Recaptcha V3 was not loaded correctly.

#### Example Usage

```yaml
ewz_recaptcha:
    version: 3 # default 2.
```

#### To Do

- [ ] Update documentation


Dynamic/Types/RecaptchaType.php
```diff
@@ -23,6 +23,11 @@ class RecaptchaType implements FormFieldTypeInterface
 {
     use SimpleTypeTrait;
 
+    public function __construct(
+        private readonly int $recaptchaVersion
```
alexander-schranz:
```suggestion
        private readonly int $recaptchaVersion = 2
```

Resources/config/type_recaptcha.xml
```diff
@@ -5,6 +5,7 @@
            xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <services>
         <service id="sulu_form.dynamic.type_recaptcha" class="Sulu\Bundle\FormBundle\Dynamic\Types\RecaptchaType">
+            <argument>%ewz_recaptcha.version%</argument>
```
alexander-schranz: This seems to fail currently in the CI

mario-fehr:
@Quehnie @alexander-schranz Does this replace #358?

Dynamic/Types/RecaptchaType.php
```diff
@@ -35,13 +49,25 @@ public function getConfiguration(): FormFieldTypeConfiguration
...
         ];
-        $builder->add($field->getKey(), \EWZ\Bundle\RecaptchaBundle\Form\Type\EWZRecaptchaType::class, $options);
+
+        if ($this->params->has('ewz_recaptcha.version')
+            && 3 == $this->params->get('ewz_recaptcha.version')
+        ) {
+            $constraint = new \EWZ\Bundle\RecaptchaBundle\Validator\Constraints\IsTrueV3();
+            $type = \EWZ\Bundle\RecaptchaBundle\Form\Type\EWZRecaptchaV3Type::class;
+            unset($options['attr']);
+        }
```
alexander-schranz, line 67: @Quehnie This should be done in a compilerpass. There you have access to the `ewz_recpatcha.version` even the bundle is registered after the form bundle: https://github.com/sulu/SuluFormBundle/blob/b01747b887e66a8dcfdfbee4332e5e329472f882/SuluFormBundle.php#L26

And then you can set a `sulu_form.recaptcha_version` parameter which you use in type_reacptacha.xml

PR #378, opened 2024-04-30
Add: Stronger type checks on user input

ghost:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

More precise checking of data sent by users.

#### Why?

We have had the problem that bots have transferred a string instead of an array. This caused the builder to run into the error `Cannot access offset of type string on string`

### DIFF
--- a/Form/Builder.php
+++ b/Form/Builder.php
@@ -100,6 +100,16 @@ class Builder implements BuilderInterface
     {
         foreach ($request->request->all() as $key => $parameters) {
             if (0 === \strpos($key, 'dynamic_')) {
+                if (!\is_array($parameters)
+                    || !\is_string($parameters['checksum'] ?? null)
+                    || !\is_string($parameters['type'] ?? null)
+                    || !\is_string($parameters['formId'] ?? null)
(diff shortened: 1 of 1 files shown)


Form/Builder.php
```diff
@@ -117,16 +127,8 @@ public function buildByRequest(Request $request): ?FormInterface
...
-                    || !isset($parameters['formId'])
-                    || !isset($parameters['formName'])
-                    || !isset($parameters['typeId'])
-                ) {
-                    continue;
-                }
-
                 $locale = $request->getLocale();
-                if (isset($parameters['locale'])) {
+                if (isset($parameters['locale']) && is_string($parameters['locale'])) {
```
alexander-schranz:
```suggestion
                if (isset($parameters['locale']) && \is_string($parameters['locale'])) {
```

alexander-schranz (commented):
Make sense, some small code style change request.

alexander-schranz:
@flogado Thank you!

PR #388, opened 2024-06-22
Fix that invalid form returns http 422 status code

### LINKED ISSUE #383: Invalid forms should return a HTTP 422 code
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | yes
| Bundle Version | 2.5
| Sulu Version | 2.5
| Browser Version | N/A

When a form isn't valid, Turbo expects a HTTP 422 code to be returned.
Otherwise form errors aren't handled correctly.

#### Actual Behavior

This bundle always returns a HTTP 200 code regardless of form errors.

#### Expected Behavior

Submit an invalid form should return a HTTP 422 when given data isn't valid.

#### Steps to Reproduce

Using Hotwire Turbo and this bundle, submit a form with invalid value and see the problem: no error displayed and no redirection.

#### Possible Solutions

Add a `kernel.response` event listener, and return a http 422 response if an invalid form is returned.
I'm working on a PR.

Jupi007:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #383
| License | MIT

#### What's in this PR?

Return the correct http code (422), when an invalid form is submitted.

#### Why?

Using Hotwire Turbo and this bundle, submit a form with invalid value is causing problems : no error displayed and no redirection.

#### To Do

- [x] Add tests (I prefer to wait your approval on the way I implemented that) ;

### DIFF
--- a/Event/RequestListener.php
+++ b/Event/RequestListener.php
@@ -17,7 +17,9 @@ use Sulu\Bundle\FormBundle\Form\BuilderInterface;
 use Sulu\Bundle\FormBundle\Form\HandlerInterface;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Event\RequestEvent;
+use Symfony\Component\HttpKernel\Event\ResponseEvent;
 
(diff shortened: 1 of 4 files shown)


alexander-schranz:
@Jupi007 Thank you!

PR #405, opened 2025-05-22
Fix for permission that allows viewing forms and data

matthieu2607:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

If a user is granted permission to view only forms and data, the Forms link does not appear in this context.

### DIFF
--- a/Admin/FormAdmin.php
+++ b/Admin/FormAdmin.php
@@ -67,7 +67,7 @@ class FormAdmin extends Admin
 
     public function configureNavigationItems(NavigationItemCollection $navigationItemCollection): void
     {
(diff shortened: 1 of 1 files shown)


alexander-schranz:
navigation points are only show if you have atleast Edit permissions: https://github.com/sulu/sulu/blob/a108d2c24082820ab1476cd2c133ca95635e12ec/src/Sulu/Bundle/CategoryBundle/Admin/CategoryAdmin.php#L44

This was a design decision to support the Selection API without having access to the menu point. So all menu points are on EDIT not on View. We will not change this in a minor version as it will have negative effects on existing projects.

Issue #409, opened 2025-07-14
Attachment URL in the admin is not working.


martinlagler:
The Attachment URL in the Data tab of the form is not correct. 
Currently it's something like this
`.../media/33/download/file`
but instead of `file` it should be the filename and version.

alexander-schranz:
The logic is currently placed here: https://github.com/sulu/SuluFormBundle/blob/b01a156d1e2f78ea5a5f546b6c678c40559ab8b9/ListBuilder/DynamicListBuilder.php#L115-L140

It would maybe be better to use the admin media redirect controller action instead of the direct download url.

Amoifr:
This one can probably be closed: it was fixed by #411 (merged in November), which switched the Data tab to the `sulu_media.redirect` route as suggested above — the PR listed #409 in its header but without a closing keyword, so GitHub kept the issue open.

PR #410, opened 2025-07-14
Fix media url in data tab

martinlagler:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | #409
| License | MIT

#### What's in this PR?

Use media redirect route to fix media url in data tab.

### DIFF
--- a/ListBuilder/DynamicListBuilder.php
+++ b/ListBuilder/DynamicListBuilder.php
@@ -51,7 +51,7 @@ class DynamicListBuilder implements DynamicListBuilderInterface
 
         foreach ($entry as $key => $value) {
             if (Dynamic::TYPE_ATTACHMENT === $dynamic->getFieldType($key)) {
-                $singleEntry[$key] = $this->getMediaUrls($value);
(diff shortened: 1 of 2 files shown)


ListBuilder/DynamicListBuilder.php
```diff
@@ -112,30 +112,13 @@ private function getMediaUrls($value): string
...
-                    'slug' => 'file',
-                    'id' => $idReplacerNumber,
-                ],
-                RouterInterface::ABSOLUTE_URL
-            ));
-        }
-
-        return $this->downloadUrl;
+        return $this->router->generate('sulu_media.redirect', [
+            'id' => $value,
```
alexander-schranz: keep the caching mechanic with `this->downloadUrl` calling router->generate is a heavy process sadly which should be avoid in this kind of loops. So we added this mechanic to only generate once a wildcard urls with `{id}` and use str_replace

PR #411, opened 2025-07-14
Fix media url in data tab

martinlagler:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | #409
| License | MIT

#### What's in this PR?

Use media redirect route to fix media url in data tab.

### DIFF
--- a/ListBuilder/DynamicListBuilder.php
+++ b/ListBuilder/DynamicListBuilder.php
@@ -51,7 +51,7 @@ class DynamicListBuilder implements DynamicListBuilderInterface
 
         foreach ($entry as $key => $value) {
             if (Dynamic::TYPE_ATTACHMENT === $dynamic->getFieldType($key)) {
-                $singleEntry[$key] = $this->getMediaUrls($value);
+                $singleEntry[$key] = $this->getMediaUrls($value, $locale);
             } else {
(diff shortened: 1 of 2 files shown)


ListBuilder/DynamicListBuilder.php
```diff
@@ -125,12 +125,13 @@ private function getDownloadUrl(): string
         if (null === $this->downloadUrl) {
             // The given id must be a number which we replace
             $idReplacerNumber = '875421';
+            $localeToReplace = 'en';
```
alexander-schranz: use `{locale}` or `__locale__` here:


```suggestion
            $localeToReplace = '{locale}';
```

The `en` is not unique enough here. If somebody change the admin url to a domain which may contains `en` letters or a directory it would replace that also. So you need something more unique like `{locale}`.

you might still need str replace url encoded version of $localeToReplace not sure if symfony router use raw url encode or url encode.

alexander-schranz: ping @martinlagler :)

martinlagler: @alexander-schranz Ups 😅  fixed now.

alexander-schranz (approved):

PR #412, opened 2025-08-22
Add sulu 3.0 support

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes none
| Related issues/PRs | non
| License | MIT

#### What's in this PR?

This PR brings FormBundle to the level of sulu 3.0

#### Why?

This will be needed when we want to use SuluFormBundle with 3.0

#### Current Known Bugs

1. Website Rendering is not working

So you have to comment out this in you template:

```
{{ form(content.form) }}
```

This can be solved in this PR later by implementing PropertyLoader.

2. Forms UI is not shown in the ADMIN

This is a strange bug as it should be there :) But for now there is no mangement UI.

#### To Do

- [ ] Fix open bugs
- [ ] Fix Testcases


Prokyonn:
See https://github.com/sulu/SuluFormBundle/pull/415

PR #413, opened 2025-09-08
Migrate rest routing to symfony routing

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| Fixed tickets | -
| Related issues/PRs | sulu/sulu#7434
| License | MIT

#### What's in this PR?
Replacing the rest routing with Symfony routing.

#### Why?
Otherwise it will break in 3.0

### DIFF
--- a/Controller/DynamicController.php
+++ b/Controller/DynamicController.php
@@ -13,7 +13,6 @@ namespace Sulu\Bundle\FormBundle\Controller;
 
 use Doctrine\ORM\EntityManager;
 use FOS\RestBundle\Controller\ControllerTrait;
-use FOS\RestBundle\Routing\ClassResourceInterface;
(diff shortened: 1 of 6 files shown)


Controller/DynamicController.php
```diff
@@ -30,7 +29,7 @@
 /**
  * Controller to create dynamic form entries list.
  */
-class DynamicController implements ClassResourceInterface
+class DynamicController
```
alexander-schranz: I don't think we removed in any bundles yet the `ClassResourceInterface` on 2.x so we should keep it here.

alexander-schranz (approved):

alexander-schranz (approved):

PR #414, opened 2025-10-23
Remove p tags from form field title

martinlagler:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | [#issuenum](https://github.com/sulu/sulu/pull/8279)
| License | MIT

#### What's in this PR?

Remove p tags from form field title to avoid save issues.

Leave overlay opens on first save, when a form field with title is existing
image

### DIFF
--- a/Form/Type/DynamicFormType.php
+++ b/Form/Type/DynamicFormType.php
@@ -95,6 +95,7 @@ class DynamicFormType extends AbstractType
             ];
 
             $title = $fieldTranslation->getTitle();
+            $title = \is_string($title) ? \str_replace(['<!--p-->', '<!--/p-->'], [''], $title) : $title;
             $placeholder = $fieldTranslation->getPlaceholder();
             $width = $field->getWidth() ?: 'full';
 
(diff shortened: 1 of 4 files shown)


Form/Type/DynamicFormType.php
```diff
@@ -94,7 +94,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'property_path' => 'data[' . $field->getKey() . ']',
             ];
 
-            $title = $fieldTranslation->getTitle();
+            $title = \str_replace(['<!--p-->', '<!--/p-->'], [''], (string) $fieldTranslation->getTitle());
```
alexander-schranz: If the value is nullable we should keep null. To avoid any bc break here.

```suggestion
            $title = $fieldTranslation->getTitle();
            $title = \is_string($title) ? \str_replace(['<!--p-->', '<!--/p-->'], [''], $title) : $title;
```

PR #415, opened 2025-11-11
Sulu3.0 integration (clean up version)

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Explain the contents of the PR.

#### Why?

Which problem does the PR fix? (remove this section if you linked an issue above)

#### Example Usage

```php
// If you added new features, show examples of how to use them here
// (remove this section if not a new feature)

$foo = new Foo();

// Now we can do
$foo->doSomething();
```

#### BC Breaks/Deprecations

#### To Do

- [ ] Create documentation


Metadata/DynamicFormMetadataLoader.php
```diff
@@ -111,11 +113,13 @@ public function warmUp($cacheDir, ?string $buildDir = null): array
 
             $fields->setDefaultType(\current($fields->getTypes())->getName());
             $section->addItem($fields);
+            /* TODO HOW?
```
alexander-schranz: no changes here still have `getItems` and `setItems` in 3.0: https://github.com/sulu/sulu/blob/9eb0f75647bd6d359ae18b6e1a91e220bb07d5ef/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php#L152

Metadata/DynamicFormMetadataLoader.php
```diff
@@ -87,10 +44,10 @@ public function __construct(
     public function warmUp($cacheDir, ?string $buildDir = null): array
     {
         $resource = __DIR__ . '/../Resources/config/forms/form_details.xml';
-        $formMetadataCollection = $this->formXmlLoader->load($resource);
-        foreach ($formMetadataCollection->getItems() as $locale => $formMetadata) {
+        $formMetadata = $this->formXmlLoader->load($resource);
+        foreach ($this->locales as $locale) {
```
alexander-schranz, line 55: this feels strange adding a section per locale looks code technical false.

alexander-schranz: its not longer metadata per locale instead one metadata containing labels with all locales that is why `$section->setLabel(string $label)` was changed to `setLabel(string $label, $locale)`. So we need where we use the translator to loop over the locales only. not to create a section per locale.

mamazu: Ah, I was wondering where the LocalizedCollection went. Fixed it.

mamazu:
@alexander-schranz Since I have no idea how the new content structure works. Can you check out the last commit and see if that is the correct implementation?

Content/PropertyResolver/SingleFormSelectionPropertyResolver.php
```diff
@@ -0,0 +1,54 @@
...
+{
+    public function __construct(private FormRepository $formRepository) {}
+
+    public function resolve(mixed $data, string $locale, array $params = []): ContentView
+    {
+        if (!\is_numeric($data)) {
+            return ContentView::create(null, ['id' => null, ...$params]);
+        }
+
+        if (!array_key_exists('resourceKey', $params)) {
```
Prokyonn: Why do we need the resourceKey here? I think we can omit this exception, as the resourceKey for the reference is always Form::RESOURCE_KEY

Content/PropertyResolver/SingleFormSelectionPropertyResolver.php
```diff
@@ -0,0 +1,54 @@
...
+            throw new \InvalidArgumentException('Missing resource key configuration on the '.self::getType().'');
+        }
+
+        /** @var string $resourceLoaderKey */
+        $resourceLoaderKey = $params['resourceLoaderKey'] ?? FormSelectionResourceLoader::getKey();
```
Prokyonn: I think we usually used `resourceLoader` without the `key`

Content/PropertyResolver/SingleFormSelectionPropertyResolver.php
```diff
@@ -0,0 +1,54 @@
...
+
+        return ContentView::createResolvableWithReferences(
+            (int) $data,
+            $resourceLoaderKey,
+            Form::RESOURCE_KEY,
+            [
+                'id' => $data,
+                'entity' => $this->formRepository->find((int)$data)?->serializeForLocale($locale),
```
Prokyonn: hmm this is a design break, as we should not use the repository in the property resolver.
But I don't think we have currently another way to adjust the `view` array

Content/ResourceLoader/FormSelectionResourceLoader.php
```diff
@@ -0,0 +1,61 @@
...
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\FormBundle\Content\ResourceLoader;
+
+use Sulu\Bundle\FormBundle\Form\BuilderInterface;
+use Sulu\Content\Application\ResourceLoader\Loader\ResourceLoaderInterface;
+use Symfony\Component\Form\FormView;
+
+final class FormSelectionResourceLoader implements ResourceLoaderInterface
```
Prokyonn: Should be only `FormResourceLoader` the resourceloader is independent of the propertyresolver and therefore should be used for both single and multi selections.

Content/PropertyResolver/SingleFormSelectionPropertyResolver.php
mamazu: From what I understand this is the type of content into which you're embedding the form (if it's an article, a snippet). Because the form needs to know how to get the title of the content it's being rendered on. (see implementation of the `TitleProviderInterface`

PR #416, opened 2025-12-11
Add Sulu 3.0 compatibility

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Adds support for Sulu 3.0 

#### To Do

- [ ] Create documentation


Tests/Functional/Metadata/DynamicFormMetadataLoaderTest.php
```diff
@@ -103,15 +103,15 @@ public function testGetMetadataGerman(): void
...
 
         $fields = $formFields->getItems()['fields'];
         $this->assertInstanceOf(FieldMetadata::class, $fields);
         $this->assertCount(28, $fields->getTypes());
         $this->assertEquals('fields', $fields->getName());
         $this->assertEquals('block', $fields->getType());
-        $this->assertEquals('attachment', $fields->getDefaultType());
-        $this->assertEquals([
+        $this->assertEquals('text', $fields->getDefaultType());
+        $this->assertEqualsCanonicalizing([
```
mamazu, line 114: This is a good point, how does 2.6 sort their State fields? Now they're no longer sorted the same way. We could just register the FormTypes with a priority attribute now to force a sorting.

Metadata/DynamicFormMetadataLoader.php
```diff
@@ -87,49 +51,49 @@ public function __construct(
...
+        $fields->setType('block');
+
+        $types = $this->formFieldTypePool->all();
+
+        $fieldTypeMetaDataCollection = [];
+        foreach ($types as $typeKey => $type) {
+            $a = $this->loadFieldTypeMetadata($typeKey, $type);
+            $fieldTypeMetaDataCollection[$a->getTitle('en')] = $a;
+        }
+        Assert::notEmpty($fieldTypeMetaDataCollection, 'No field type metadata loaded');
```
mamazu, line 68: We're using `addType` that's why we don't need to generate a key explicitly in the array.

```suggestion
        $fieldTypeMetaDataCollection = [];
        foreach ($types as $typeKey => $type) {
            $fieldTypeMetaDataCollection[] = $this->loadFieldTypeMetadata($typeKey, $type);
        }
        Assert::notEmpty($fieldTypeMetaDataCollection, 'No field type metadata loaded');
```

mamazu (approved):

TitleProvider/TitleProviderPool.php
```diff
@@ -17,18 +17,16 @@
...
-     * TitleProviderPool constructor.
-     *
-     * @param TitleProviderInterface[] $providers
+     * @param iterable<string, TitleProviderInterface> $providers
      */
-    public function __construct(array $providers)
+    public function __construct(iterable $providers)
     {
-        $this->providers = $providers;
+        $this->providers = $providers instanceof \Traversable ? \iterator_to_array($providers) : $providers;
```
mamazu, line 29: This is how we handle it in the core.

```suggestion
        $this->providers = [...$providers];
```

alexander-schranz (approved):

alexander-schranz:
May some services should be better be internal. But lets merge this and create a first RC release.

alexander-schranz:
Thx @Prokyonn @mamazu 👍

PR #418, opened 2026-02-03
Fix metadata loading of form_details

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fix metadata loading ofform_details.

#### Why?

This is a quickfix which fixes a problem with all form metadata loading when FormBundle is installed: https://github.com/sulu/sulu/pull/8595

To avoid such problems we should migrate


Metadata/DynamicFormMetadataLoader.php
```diff
@@ -90,6 +93,10 @@ public function warmUp($cacheDir, ?string $buildDir = null): array
 
     public function getMetadata(string $key, string $locale, array $metadataOptions = []): ?MetadataInterface
     {
+        if ('form_details' !== $key) {
+            return null;
+        }
```
alexander-schranz, line 98: as discussed already with @Prokyonn every custom form metadata loader need to early return as soon as possible. Still its open to migrate this whole loader to a visitor instead. Currently mark it as `@internal` so we can tackle that at any time.

Prokyonn (approved):

chirimoya:
@alexander-schranz please also check the doctrine dependencies.

Event/CacheInvalidationListener.php
```diff
@@ -33,12 +32,12 @@ public function __construct(?CacheManagerInterface $cacheManager)
         $this->cacheManager = $cacheManager;
     }
 
-    public function postUpdate(PostUpdateEventArgs|LifecycleEventArgs $eventArgs): void
+    public function postUpdate(PostUpdateEventArgs $eventArgs): void
```
alexander-schranz, line 35: @Prokyonn the `Doctrine\ORM\Event\LifecycleEventArgs` does not exist in this case is it still required?

Prokyonn: As long as we do not need it in `lowest` then no

Prokyonn (approved):

composer.json
```diff
@@ -18,9 +18,9 @@
     ],
     "require": {
         "php": "^8.2",
-        "doctrine/collections": "^2.0",
+        "doctrine/collections": "^1.6 || ^2.0",
         "doctrine/orm": "^2.17.3 || ^3.3",
-        "doctrine/persistence": "^3.0",
+        "doctrine/persistence": "^3.1 || ^4.0",
         "sulu/sulu": "^3.0",
```
alexander-schranz:
```suggestion
        "sulu/sulu": "^3.0.4",
```

alexander-schranz: this changes would else break compatibility from: 3.0.0-3.0.3

composer.json
```diff
@@ -18,10 +18,10 @@
     ],
     "require": {
         "php": "^8.2",
-        "doctrine/collections": "^2.0",
+        "doctrine/collections": "^1.6 || ^2.0",
         "doctrine/orm": "^2.17.3 || ^3.3",
-        "doctrine/persistence": "^3.0",
-        "sulu/sulu": "^3.0",
+        "doctrine/persistence": "^3.1 || ^4.0",
+        "sulu/sulu": "^3.0.4",
```
alexander-schranz:
```suggestion
        "sulu/sulu": "^3.0.4 || 3.0.x-dev",
```

PR #423, opened 2026-02-18
Replace $request->get with correct methods for Symfony 8 compatibility

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT

#### What's in this PR?
Preparation for Symfony 8

#### Why?
Do not use deprecated functions


Form/Builder.php
```diff
@@ -295,7 +295,7 @@ private function getWebspaceKey(): ?string
             return null;
         }
 
-        $suluAttributes = $request->get('_sulu');
+        $suluAttributes = $request->attributes->get('_sulu');
```
mamazu, line 298: This entire method could be replaced with the `RequestAnalyzer` in case the webspace key resolution changes.

Tests/Application/config/config.yml
```diff
@@ -10,6 +10,7 @@ framework:
 
 doctrine:
     orm:
+        enable_nativ_lazy_objects: true
```
mamazu: This is required because if you upgrade the plugin to the newest version it already installs the Symfony 8 Vardumper which needs this feature.

mamazu:
@alexander-schranz Can you review this again then the bundle should be symfony 8 ready as well.

alexander-schranz:
@mamazu thank you!

alexander-schranz:
@mamazu next step would be upgrade composer.json symfony and doctrine dependencies.

PR #431, opened 2026-05-19
Add shadow locale fallback for form resource loader and twig extension

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

The `FormResourceLoader` now falls back to the shadow locale passed via the `_shadowLocale` parameter when a form has no translation for the requested locale. It also filters loaded forms by the actual translation, since `loadByIds` does not respect the locale on its own. The `FormTwigExtension` receives the request analyzer and resolves a missing locale either from the current localization or from the form's default locale.

#### Why?

Forms rendered on shadow pages stayed empty because the loader returned no result when the form had no translation in the shadow's target locale. Templates that called the form twig function without a locale could also end up with `null`, which broke the rendered form. The fallback chain restores the expected output for shadow locales and locale-less twig calls.

### DIFF
--- a/Twig/FormTwigExtension.php
+++ b/Twig/FormTwigExtension.php
@@ -13,23 +13,17 @@ namespace Sulu\Bundle\FormBundle\Twig;
 
 use Sulu\Bundle\FormBundle\Entity\Form;
 use Sulu\Bundle\FormBundle\Form\BuilderInterface;
+use Sulu\Component\Webspace\Analyzer\RequestAnalyzerInterface;
 use Symfony\Component\Form\FormView;
 use Twig\Extension\AbstractExtension;
 use Twig\TwigFunction;
 
-/**
- * Extension for content form generation.
- */
 class FormTwigExtension extends AbstractExtension
 {
-    /**
-     * @var BuilderInterface
-     */
-    private $formBuilder;
-
-    public function __construct(BuilderInterface $formBuilder)
-    {
-        $this->formBuilder = $formBuilder;
+    public function __construct(
+        private BuilderInterface $formBuilder,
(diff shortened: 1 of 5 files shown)


alexander-schranz (approved):

Twig/FormTwigExtension.php
```diff
@@ -66,6 +60,13 @@ public function getFormByContent(array $formContent, string $type, string $typeI
             return null;
         }
 
+        if (null === $locale) {
+            $requestLocale = $this->requestAnalyzer->getCurrentLocalization()?->getLocale();
+            $locale = (null !== $requestLocale && null !== $form->getTranslation($requestLocale))
+                ? $requestLocale
+                : $form->getDefaultLocale();
```
alexander-schranz: this looks unexpected getDefaultLocale is normally something only for the admin not used by the website.

Shouldn't this be more shadowLocale with fallback to requestLocale?

Prokyonn: Hmm yes I agree, but I don't think we have easy access to the shadowLocale in the extension 🤔

alexander-schranz: we have it in twig? then we can give it as a own param into the extension: https://github.com/sulu/skeleton/blob/fcca456b4c103b16728d1f978306f285e92c65bc/templates/base.html.twig#L13

Prokyonn: Refactored to use the object from the requestAttributes

PR #434, opened 2026-06-10
Remove deprecated 2.x APIs and features for Sulu 3.0

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

All code that was deprecated in the 2.x line has been removed. This includes the static forms feature (`AbstractType`, `TypeInterface`, `FormConfigurationFactory::buildByType()`, `ListController`, `ListProviderInterface`/`ListProviderRegistry` and their routes), the `DynFormSavedEvent`, the `HandlerInterface` event name constants, Swiftmailer support (`Mail\Helper`, `Mail\NullHelper`, `swift_mailer.xml`), the `dynamic.html.twig` form theme, and the top-level `sulu_form.media_collection_strategy` configuration option. The deprecated ESI-based CSRF token loading and the `max` attribute on attachment fields have also been removed. All breaking changes are documented in `UPGRADE-3.0.md`.

#### Why?

Sulu 3.0 is a new major version and the right moment to drop all the deprecated shims that accumulated in the 2.x series. Removing them reduces the bundle's surface area, eliminates dead code paths that had to be maintained, and ensures consumers get a clean compile-time error instead of a silent deprecation warning when they rely on removed APIs.


martinlagler (approved):

Resources/config/services.xml
```diff
@@ -9,7 +9,7 @@
         <!-- Form Handler -->
         <service id="sulu_form.handler" class="Sulu\Bundle\FormBundle\Form\Handler">
             <argument type="service" id="doctrine.orm.entity_manager" />
-            <argument type="service" id="sulu.mail.helper" />
+            <argument type="service" id="Sulu\Bundle\FormBundle\Mail\HelperInterface" />
```
alexander-schranz: bundles never should use `classes` or `interfaces` as service definitions.

Resources/config/mailer.xml
```diff
@@ -10,5 +10,6 @@
             <argument>%sulu_form.mail.sender%</argument>
             <argument type="service" id="logger" />
         </service>
+        <service id="Sulu\Bundle\FormBundle\Mail\HelperInterface" alias="sulu.mail.mailer" />
```
alexander-schranz: would not expose this service as a Interface

Resources/doc/index.md
```diff
@@ -37,12 +37,12 @@ sulu_form:
...
 
-Optional configure the email handler to SwiftMailer (`swift_mailer`) or the Symfony Mailer (`mailer`):
+Optionally configure the email handler (defaults to the Symfony Mailer):
 
 ```yml
 sulu_form:
     mail:
-        helper: "swift_mailer" # is default
+        helper: "mailer" # is default
 ```
```
alexander-schranz: We should remove this config as Swiftmailer is not longer supported, null helper removed and so there is no switch between different transports yet.

alexander-schranz (approved):

PR #435, opened 2026-06-10
Restore 2.6 form rendering for the single form selection content type

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | yes
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT
| Documentation PR | sulu/sulu-docs# 

#### What's in this PR?

The single form selection content type again resolves to a ready-to-render `FormView`, so existing 2.6 templates render the form unchanged without the previously required `sulu_form_build` Twig call. The `FormResourceLoader` now builds the view by reading the current content object from the request and resolving its plural resourceKey. Form title providers and the value stored in `fo_dynamics.type` switched from the singular content type key to the plural resourceKey (`pages`, `articles`, `snippets`). A Doctrine migration converts existing submission `type` values to the plural keys, and the unit and content resolution tests were expanded to cover the loader, resolver and migration.

#### Why?

In the Sulu 3.0 content resolution rewrite the form was no longer built during content resolution, which broke template parity with 2.6 and forced every project to update its form templates. Building the view in the resource loader restores the original rendering behavior. Moving to the plural resourceKey aligns the form type with the rest of Sulu 3.0, and the migration keeps existing submissions consistent after the upgrade.

PR #438, opened 2026-06-10
Harden type declarations across

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | yes
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Native parameter and return types, generic PHPDoc annotations and nullability guards were added across the bundle to satisfy stricter static analysis. The phpstan baseline shrank from 2204 to 392 ignored entries and the suite now passes clean. Several public signatures were tightened, including `Dynamic::getForm()` which now returns `?Form`, `Mail\HelperInterface::sendMail()` whose parameters gained native types, and the controller helper methods which gained explicit return types.

The change also fixes `Dynamic::getDate()` which previously read the wrong field due to a typo. All breaking changes are listed in `UPGRADE-3.0.md`.

#### Why?

Tightening the types removes that suppression and catches nullability and shape mismatches at analysis time instead of at runtime. The `getDate()` typo is corrected in the same pass since it surfaced while hardening the entity accessors.


Command/FormGeneratorCommand.php
```diff
@@ -56,6 +56,9 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 
         foreach ($locales as $locale) {
             $formTranslation = $form->getTranslation($locale, true);
+            if (null === $formTranslation) {
+                throw new \RuntimeException(\sprintf('Could not create form translation for locale "%s".', $locale));
+            }
```
alexander-schranz: unnecessary with second parameter true this always returns a FormTranslation.

You could use phpstan conditional return types:

```
     * @return ($create is true ? FormTranslation : ?FormTranslation)
```

https://phpstan.org/blog/phpstan-1-6-0-with-conditional-return-types

Configuration/FormConfigurationFactory.php
```diff
@@ -63,31 +65,35 @@ public function __construct(
      */
     public function buildByDynamic(Dynamic $dynamic): FormConfigurationInterface
     {
+        // The form can be null for orphaned submissions (the formId column is set to null
+        // via "on-delete=SET NULL" when the form is deleted). A configuration cannot be built then.
         $form = $dynamic->getForm();
+        if (null === $form) {
+            throw new \RuntimeException('The given dynamic submission is not related to a form anymore.');
+        }
```
alexander-schranz: @chirimoya we did not remove the data currently if the form got removed. Should we change this in 3.0? Currently form submissions are never removed and stay forever.

chirimoya: Yes.

alexander-schranz: @Prokyonn we can change then the relation to `on-cascade="DELETE"` so its never nullable. We need migration which removes all orphan entities of Dynamic.

Configuration/FormConfigurationFactory.php
```diff
@@ -97,7 +103,7 @@ private function buildAdminMailConfigurationByDynamic(Dynamic $dynamic): ?MailCo
 
         $adminMailConfiguration->setSubject($translation->getSubject());
         $adminMailConfiguration->setFrom(
-            $this->getEmail($translation->getFromEmail(), $translation->getFromName())
+            $this->getEmail($translation->getFromEmail(), $translation->getFromName()) ?: []
```
alexander-schranz, line 101: did you test if no email is defined that the email is still send to the configured email in `sulu_form` configs?

alexander-schranz: Because I think previously even it was not typed that way `to` was able to be null in that case so `??` did work?

For me the empty array seems more confusing as go with `null` in the MailConfiguration.

alexander-schranz: okay seems the helper can work with '', [], as it uses `?:` instead of `??` here: https://github.com/sulu/SuluFormBundle/blob/7d2a3ac229eff4f1ac56270b95857ca83cf499fc/Mail/MailerHelper.php#L79

Still make sure to test if it works expected.

UPGRADE-3.0.md
```diff
@@ -2,6 +2,55 @@
...
+The relation is now `on-delete=CASCADE` and the `formId` column is `NOT NULL`. Deleting a form now also deletes all of
+its submissions. Accordingly, **`Dynamic::getForm()` keeps returning a non-nullable `Form`** and the `Dynamic`
+constructor now requires a non-null `Form` as its fourth argument.
+
+Because existing databases may still contain orphaned submissions, you must remove them before applying the new schema
+(the `NOT NULL` constraint cannot be added while `NULL` values exist). Generate the schema migration with
+`bin/console doctrine:migrations:diff` and prepend the orphan cleanup so it runs first:
+
+```php
+public function up(Schema $schema): void
```
Prokyonn: TODO add a doctrine migration as soon as https://github.com/sulu/SuluFormBundle/pull/435 is merged

wachterjohannes (approved):

PR #442, opened 2026-08-24
Keep the fo_dynamics audit foreign keys when changing the form relation

### LINKED ISSUE #441: Version20260702120000 drops the creator/changer foreign keys on fo_dynamics
## Summary

`Version20260702120000` drops **every** foreign key on `fo_dynamics` and only recreates the one targeting `fo_forms`. The `creator` / `changer` foreign keys towards `se_users` are removed as a side effect and never restored, neither by `up()` nor by `down()`.

## Where it happens

`migrations/Version20260702120000.php`:

```php
private function changeFormRelation(Schema $schema, bool $notnull, string $onDelete): void
{
    $table = $schema->getTable(self::TABLE);
    $newTable = clone $table;

    foreach ($newTable->getForeignKeys() as $foreignKey) {
        $newTable->removeForeignKey($foreignKey->getName());   // <-- removes all of them
    }

    $newTable->getColumn('formId')->setNotnull($notnull);
    $newTable->addForeignKeyConstraint(self::FORM_TABLE, ['formId'], ['id'], ['onDelete' => $onDelete]);

    $diff = $this->sm->createComparator()->compareTables($table, $newTable);

    foreach ($this->platform->getAlterTableSQL($diff) as $sql) {
        $this->connection->executeStatement($sql);
    }
}
```

`$schema` is the live database schema, so the loop also picks up the two audit foreign keys.

The migration's own description states the intent is narrower: *"make the fo_dynamics form relation non-nullable with ON DELETE CASCADE"*.

## Why those foreign keys exist

`Sulu\Bundle\FormBundle\Entity\Dynamic` uses `AuditableTrait` and implements `AuditableInterface`, so Sulu's `UserBlameSubscriber` maps `creator` and `changer` at runtime on `loadClassMetadata`, with join columns `idUsersCreator` / `idUsersChanger` and `onDelete: SET NULL`. They are part of the expected schema even though they appear in no mapping file.

## Impact

After running the migration, `fo_dynamics` keeps the `idUsersCreator` / `idUsersChanger` columns and their indexes but loses the constraints, so `ON DELETE SET NULL` no longer applies: deleting a user leaves dangling user ids in `fo_dynamics`.

`doctrine:migrations:diff` then keeps proposing to add them back on every run, which is how I found it.

## Reproduction

On a PostgreSQL 16 / Sulu 3.0 / SuluFormBundle 3.0.0 install:

1. Let the schema be created with the audit foreign keys present (they are, on every other auditable table).
2. Run the bundle migrations.
3. Compare:

```sql
SELECT c.table_name,
       (SELECT count(*) FROM pg_constraint pc
         WHERE pc.conrelid = (quote_ident(c.table_name))::regclass
           AND pc.contype = 'f'
           AND pg_get_constraintdef(pc.oid) ILIKE '%se_users%') AS user_fks
FROM information_schema.columns c
WHERE c.column_name IN ('iduserscreator', 'iduserschanger')
GROUP BY c.table_name ORDER BY user_fks;
```

On my install `fo_dynamics` is the only table out of 21 reporting `0`; every other auditable table reports `2`, including `fo_form_translations` from this same bundle.

## Suggested fix

Only drop the foreign key that is actually being changed:

```php
foreach ($newTable->getForeignKeys() as $foreignKey) {
    if (0 === strcasecmp($foreignKey->getForeignTableName(), self::FORM_TABLE)) {
        $newTable->removeForeignKey($foreignKey->getName());
    }
}
```

Installations that already ran the migration would still need a follow-up migration to restore the two constraints, since `down()` does not bring them back either.

steeven-th:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #441
| License | MIT

#### What's in this PR?

`Version20260702120000` removed **every** foreign key of `fo_dynamics` before recreating only the one targeting `fo_forms`. The `creator` / `changer` constraints, mapped at runtime by `UserBlameSubscriber`, were therefore dropped and never restored, neither by `up()` nor by `down()`.

This PR contains three things:

- The fix: only the foreign key targeting `fo_forms` is removed now, the others are preserved.
- A follow-up migration, `Version20260824120000`, restoring the two `se_users` constraints on installations that already ran the faulty one. It goes through the Doctrine Schema API, skips constraints that are already there, and does nothing when the `se_users` table is absent.
- The tests: an assertion added to `Version20260702120000Test` checking the audit foreign keys survive, and a `Version20260824120000Test` covering both the restore and a second no-op run.

#### Why?

After the migration ran, `fo_dynamics` kept its `idUsersCreator` / `idUsersChanger` columns and their indexes but lost the constraints, so `ON DELETE SET NULL` no longer applied: deleting a user left dangling user ids behind.

The visible side effect, and how I found it, is that `doctrine:migrations:diff` kept proposing to add those two keys back on every run. On my installation `fo_dynamics` was the only auditable table out of 21 without them, while `fo_form_translations` from this same bundle had them.

#### Tests

Checked on **PostgreSQL 14 and MySQL 8** with PHP 8.4, matching the CI matrix.

- Fresh schema, original migration: only one foreign key left, bug reproduced.
- Fresh schema, fixed migration: all three foreign keys kept.
- Already broken database, follow-up migration: all three restored.
- Follow-up run a second time: no change, no error.

The new test **fails on the original code** (`Failed asserting that an array contains 'iduserscreator'`) and passes with the fix.

`php-cs-fixer` and `phpstan` are clean on the changed files.

#### Note

`se_users` is hardcoded in the follow-up migration, since that table name is fixed in the `Sulu\Bundle\SecurityBundle\Entity\User` mapping. A guard on the table's existence keeps the migration from failing on an installation that departs from it. Happy to take another approach if you prefer.

### DIFF
--- a/Tests/Functional/Migrations/Version20260702120000Test.php
+++ b/Tests/Functional/Migrations/Version20260702120000Test.php
@@ -68,6 +68,16 @@ class Version20260702120000Test extends SuluTestCase
         self::assertSame(0, $this->countDynamic('linked'));
     }
 
+    public function testUpKeepsTheAuditForeignKeys(): void
+    {
+        $this->createMigration()->up($this->introspectSchema());
+
+        $localColumns = $this->foreignKeyColumns();
+
+        self::assertContains('iduserscreator', $localColumns);
+        self::assertContains('iduserschanger', $localColumns);
+    }
+
     private function createMigration(): Version20260702120000
     {
         return new Version20260702120000($this->connection, new NullLogger());
@@ -76,6 +86,22 @@ class Version20260702120000Test extends SuluTestCase
     private function introspectSchema(): Schema
     {
         return $this->connection->createSchemaManager()->introspectSchema();
+    }
+
+    /**
+     * @return list<string>
+     */
+    private function foreignKeyColumns(): array
+    {
+        $columns = [];
+
+        foreach ($this->introspectSchema()->getTable('fo_dynamics')->getForeignKeys() as $foreignKey) {
+            foreach ($foreignKey->getLocalColumns() as $localColumn) {
+                $columns[] = \strtolower($localColumn);
+            }
+        }
+
+        return $columns;
     }
 
     private function insertForm(): int
--- a/Tests/Functional/Migrations/Version20260824120000Test.php
+++ b/Tests/Functional/Migrations/Version20260824120000Test.php
@@ -1,0 +1,92 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\FormBundle\Tests\Functional\Migrations;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Schema\Schema;
+use Psr\Log\NullLogger;
+use Sulu\Bundle\FormBundle\Migrations\Version20260824120000;
+use Sulu\Bundle\TestBundle\Testing\SuluTestCase;
+
+class Version20260824120000Test extends SuluTestCase
+{
+    private Connection $connection;
+
+    protected function setUp(): void
+    {
+        self::bootKernel();
+        self::purgeDatabase();
+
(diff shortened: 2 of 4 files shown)


steeven-th:
Opened as promised in #441, @Amoifr.

PR #443, opened 2026-08-27
Allow Symfony 8

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT

#### What's in this PR?
Allow installing Symfony 8 packages.

#### Why?
Test if any other breaking changes need to be addressed.


alexander-schranz:
One blocker:

> excelwebzone/recaptcha-bundle v1.5.42 requires symfony/form (^2.8 || ^3.0 || ^4.0 || ^5.0 || ^6.0 || ^7.0)

Amoifr:
Ran your branch locally to see where it stands. The blocker has moved since Alexander's comment: composer now resolves fine on 8.4 and 8.5, and the three red jobs fail later, at `Bootstrap test environment`, with

```
Attempted to load class "XmlFileLoader" from namespace "Symfony\Component\DependencyInjection\Loader"
in ./DependencyInjection/SuluFormExtension.php
```

Symfony 8 removed the DI XML loader, so the ten files under `Resources/config` that `SuluFormExtension` loads (501 lines, several of them conditional) have to move to the PHP `ContainerConfigurator` format. Two more fatals wait behind it, both from interfaces that gained a return type in Symfony 8: `DynamicFormType::getBlockPrefix()` needs `: string`, and `MailchimpListSubscriber` and `ProtectedMediaSubscriber` need `getSubscribedEvents(): array`. The `Request::get()` side is already done, #423 covered it.

On `excelwebzone/recaptcha-bundle`: worth knowing that the upstream repository is **archived** since February, so it will never get Symfony 8. The good news is that it is only a `require-dev` plus `suggest` here, so it blocks the test suite rather than anyone installing this bundle on Symfony 8. If you want a replacement rather than dropping the dev dependency, `beelab/recaptcha2-bundle` (`^6.4 || ^7.4 || ^8.0`), `karser/karser-recaptcha3-bundle` and `victor-prdh/recaptcha-bundle` all support Symfony 8 today.

I have done that XML to PHP conversion on five sibling bundles this week (theme and phpcr-migration are merged, comment, headless and community are open), so I am happy to take it on if that helps: either as a PR against your branch, or as a separate one you rebase on. Just say which you prefer, I do not want to step on your work.

mamazu:
Thank you for your in depth analysis. I've fixed the two missing return types you mentioned in this PR and split out another PR for the PHP service migration. (see #444 )

As for the recaptcha implementation, that's a good question I don't have any idea which is the more popular/trusted one to replace the current implementation. But I guess as long as they still implement the same interfaces, it shouldn't really matter.

draconivis:
I tested this branch and found a few more spots that still need fixing, on top of what's already fixed here and split into #444:

`Form/Type/DynamicFormType.php` — two more methods need Symfony 8's now-typed `AbstractType/FormTypeInterface` signatures (same category as the `getBlockPrefix()` fix already in this PR):
```diff
- public function buildForm(FormBuilderInterface $builder, array $options)
+ public function buildForm(FormBuilderInterface $builder, array $options): void
- public function configureOptions(OptionsResolver $resolver)
+ public function configureOptions(OptionsResolver $resolver): void
```

`Dynamic/Types/AttachmentType.php` — three validator constraints still use the legacy array-options constructor, which Symfony 8's validator removed entirely:
```diff
- $fileConstraint = new File(['mimeTypes' => $mimeTypes]);
+ $fileConstraint = new File(mimeTypes: $mimeTypes);

- $constraints[] = new All(['constraints' => $allConstraints]);
+ $constraints[] = new All(constraints: $allConstraints);

- $constraints[] = new Count([
-     'max' => $fileMax,
- ]);
+ $constraints[] = new Count(max: $fileMax);
```
`File` crashed with `InvalidArgumentException`, `Count` with `TypeError`, `All` with `ConstraintDefinitionException`.
 
`Tests/Application/templates/overview.html.twig` — `app.request.get('send')` on line 10. `Request::get()` was removed in Symfony 8 too, not just the PHP-side call sites #423 already covered. Should probably use `app.request.query.get('send')` here as well.

With all of the above fixed locally, got a full page render + form submission (including a file upload) working end to end on Symfony 8. Didn't hit the XmlFileLoader failure from #444 in this setup since my dependency resolution kept symfony/dependency-injection/framework-bundle on 7.4.x — no 8.0 release of those exists yet — so that one's still open separately, this isn't a report that it's fixed.

alexander-schranz:
The Service definition changes are merged. Can you rebase this one.

Tests/Application/config/reference.php
draconivis, line 1: add this file to the gitignore

draconivis (changes requested):
please fix lint and remove the `reference.php`. else it looks good!

Amoifr:
Same trap I hit on the other bundles of sulu/sulu#8216, so here is what worked there.

The gitignore alone is not enough: from Symfony 8.1 on, FrameworkBundle regenerates `reference.php` on every boot of the test kernel, into a directory git tracks. It comes back for everyone, and it stays visible to php-cs-fixer and to PHPStan. Three one-line entries cover it:

```
# .gitignore
Tests/Application/config/reference.php
```

```php
// .php-cs-fixer.dist.php, on the finder
->notPath('Tests/Application/config/reference.php')
```

```yaml
# phpstan.neon, under excludePaths
- '%currentWorkingDirectory%/Tests/Application/config/reference.php (?)'
```

The `(?)` suffix on the PHPStan path is the part that is easy to miss: without it the analysis fails whenever the file has not been generated yet.

One more thing, the lint job is red on two files rather than one:

```
1) Dynamic/Types/AttachmentType.php (native_function_invocation)
2) Tests/Application/config/reference.php (header_comment)
```

Happy to send this as a PR against your branch if that is easier.

mamazu:
> One blocker:
> 
> > excelwebzone/recaptcha-bundle v1.5.42 requires symfony/form (^2.8 || ^3.0 || ^4.0 || ^5.0 || ^6.0 || ^7.0)

@alexander-schranz What do we want to do about it. @Amoifr suggested a list of replacement packages. Do any of them look good enough to be recommended and installed in the stack?

Tests/Application/config/reference.php
mamazu: Fixed.

alexander-schranz:
For this pull request, remove `excelwebzone/recaptcha-bundle` from the `composer.json` but keep all code as it is and install the package only on the CI via `composer require excelwebzone/recaptcha-bundle --no-update` for the versions running against <=Symfony 7.4 (think PHP <= 8.3) for backwards compatibility reasons.

I think such bridges should live in its own packages in future and core should stay independent and being blocked by such packages. We are more about removing more and more dependencies over add new once to core.

Also the question is if a Content Manager even should care about Captchas or if the developer just install directly a recaptcha bundle of there choice and add it via Symfony Form Extensions to the Dynamic form. An upcoming pull request could deprecate the captcha implementation and more go with a documentation (even for example for Captcha which provide only none Symfony Bundle), but need discuss this with the other core team members.

.github/workflows/test-application.yaml
```diff
@@ -112,6 +112,10 @@ jobs:
               # testing lowest versions.
               run: composer remove "*php-cs-fixer*" "*phpstan*" "*rector*" --dev --no-update
 
+            - name: Install recaptcha bundle for backwards compatibility
+              if: ${{ matrix.php-version }} == "8.2" || ${{ matrix.php-version }} == "8.3"
+              run: composer config minimum-stability ${{ matrix.composer-stability }}
```
draconivis, line 117: this just runs the same command as the next step?


```suggestion
            - name: Install recaptcha bundle for backwards compatibility
              if: ${{ matrix.php-version }} == "8.2" || ${{ matrix.php-version }} == "8.3"
              run: composer require excelwebzone/recaptcha-bundle:^1.4.2 --no-update
```

Tests/Functional/Metadata/DynamicFormMetadataLoaderTest.php
draconivis, line 1: this test will fail when the CI actually installs recaptcha again.

i guess you'd have to check for it in the test and then conditionally test it. for example:
```php
$expected = [
    'attachment', 'checkbox', /* ... the 27 ... */ 'zip',
];
if (\class_exists(\EWZ\Bundle\RecaptchaBundle\Form\Type\EWZRecaptchaType::class)) {
    $expected[] = 'recaptcha';
}
$this->assertEqualsCanonicalizing($expected, \array_keys($fields->getTypes()));
```
and for the attachment tests:
```php
$this->assertCount(
    \class_exists(\EWZ\Bundle\RecaptchaBundle\Form\Type\EWZRecaptchaType::class) ? 28 : 27,
    $fields->getTypes()
);
```

draconivis (changes requested):

Amoifr:
Confirming draconivis's first point, since the evidence is a bit hidden: the step is named "Install recaptcha bundle for backwards compatibility" but its `run:` is `composer config minimum-stability ${{ matrix.composer-stability }}`, which is the next step copied. So it never installs the bundle today, and that is why CI is green right now: the tests draconivis expects to break are not reached yet. They will be as soon as the `run:` is fixed, so the conditional assertions he suggests are needed in the same commit.

One thing worth double checking in the suggested block: GitHub expressions use single quotes for strings, so `if: matrix.php-version == '8.2' || matrix.php-version == '8.3'` is likely what you want, without the `${{ }}` wrapper since `if:` is already an expression context. No other workflow in sulu compares a matrix value, they all use the bare `if: ${{ matrix.something }}` form, so there is no precedent to copy from here.

draconivis (approved):

Prokyonn:
symfony/form and symfony/yaml installs only 7.4 in the highest CI run

.github/workflows/test-application.yaml
```diff
@@ -112,6 +112,10 @@ jobs:
               # testing lowest versions.
               run: composer remove "*php-cs-fixer*" "*phpstan*" "*rector*" --dev --no-update
 
+            - name: Install recaptcha bundle for backwards compatibility
+              if: ${{ matrix.php-version }} == "8.2" || ${{ matrix.php-version }} == "8.3"
```
draconivis:
```suggestion
              if: ${{ matrix.php-version == '8.2' || matrix.php-version == '8.3' }}
```

draconivis:
@Prokyonn fixed

Prokyonn (approved):

Prokyonn:
@mamazu Thank you! :)

PR #444, opened 2026-08-28
Migrating from XML to PHP service definitons

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | technically but you shouldn't be depending on 
| Deprecations? | no
| License | MIT

#### What's in this PR?

Replacing service xml with php

#### Why?

XML is no longer supported in Symfony 8


DependencyInjection/SuluFormExtension.php
```diff
@@ -360,7 +359,7 @@ public function load(array $configs, ContainerBuilder $container): void
         }
 
         if ($config['media']['protected']) {
-            $loader->load('protected_media.xml');
+            $loader->load('services_media.php');
```
alexander-schranz, line 362: this should be named `protected_media.php`. it has nothing todo if media bundle is here or not. okay to prefix it with services, but media is not enough.

alexander-schranz (changes requested):
Small change request. Can we do the same for 2.x branch so backmerging if any issues appear is easy.

mamazu:
I've created a seperate PR and will Rebase this onto the other one when it's merged.

PR #445, opened 2026-08-31
Migrating from XML to PHP service definitons

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?
Migrating services to php

#### Why?
Symfony 8


composer.json
```diff
@@ -21,6 +21,7 @@
         "doctrine/collections": "^1.0 || ^2.0",
         "doctrine/orm": "^2.5.3",
         "doctrine/persistence": "^1.3 || ^2.0 || ^3.0",
+        "gromnan/symfony-config-xml-to-php": "^0.2.1",
```
alexander-schranz: remove this dependency :)

mamazu:
Gone

Resources/config/mailer.php
draconivis, line 1: This file is now an empty no-op — the `sulu.mail.mailer` service it used to hold moved to `services.php:68`. `services.php` is loaded unconditionally, but this file is still loaded only when `interface_exists(MailerInterface::class)` (`SuluFormExtension.php:279-282`). That guard no longer gates this service at all. Suggest moving the definition back here, or dropping the now-pointless conditional load.

Resources/config/services.php
```diff
@@ -0,0 +1,280 @@
...
+        ->args([new Reference('logger')]);
+
+    $services->set('sulu.mail.mailer', MailerHelper::class)
+        ->args([
+            new Reference('mailer.mailer'),
+            '%sulu_form.mail.from%',
+            '%sulu_form.mail.to%',
+            '%sulu_form.mail.sender%',
+            new Reference('logger'),
+        ]);
```
draconivis: This service used to live in the conditionally-loaded `mailer.xml` — see the comment on `mailer.php`

Resources/config/protected_media.php
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Symfony\Component\DependencyInjection\Loader\Configurator;
+
+return static function(ContainerConfigurator $container) {
+    $services = $container->services();
+    $parameters = $container->parameters();
+
+    $services->set('sulu_form.private_media_subscriber', \Sulu\Bundle\FormBundle\Event\ProtectedMediaSubscriber::class)
+        ->args([
+            service('router'),
```
draconivis: Uses the `service()` helper here, while `services.php`, `services_article.php`, `swift_mailer.php`, `title-providers.php` and `type_sendinblue.php` all use `new Reference(...)` for the same thing. `services_trash.php` has the same inconsistency

Resources/config/services_trash.php
```diff
@@ -0,0 +1,28 @@
...
+return static function(ContainerConfigurator $container) {
+    $services = $container->services();
+    $parameters = $container->parameters();
+
+    $services->set('sulu_form.form_trash_item_handler', \Sulu\Bundle\FormBundle\Trash\FormTrashItemHandler::class)
+        ->args([
+            service('sulu_trash.trash_item_repository'),
+            service('doctrine.orm.entity_manager'),
+            service('sulu_trash.doctrine_restore_helper'),
+            service('sulu_activity.domain_event_collector'),
```
draconivis: Same `service()` vs `new Reference(...)` inconsistency as `protected_media.php` — the rest of the converted files use `new Reference(...)`.

Resources/config/title-providers.php
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Symfony\Component\DependencyInjection\Loader\Configurator;
+
+use Sulu\Bundle\FormBundle\TitleProvider\StructureTitleProvider;
+use Sulu\Bundle\FormBundle\TitleProvider\TitleProviderPool;
+use Symfony\Component\DependencyInjection\Reference;
+
+return static function(ContainerConfigurator $container) {
+    $services = $container->services();
+    $parameters = $container->parameters();
```
draconivis: `$parameters = $container->parameters();` is declared but never used. Same dead line in `mailer.php:16`, `protected_media.php:16`, `services_trash.php:16`, `type_mailchimp.php:20`, `type_sendinblue.php:22`.

Resources/config/services_article.php
draconivis, line 1: This renames `article.xml` → `services_article.php`, intentional?

Resources/config/services.php
mamazu: Fixed. Apparently this was only unconditional in 3.0

Resources/config/services_article.php
mamazu: Yes, the idea for the naming convention was that every external bundle that we integrate has the `service_` prefix and then the name of the bundle and the article bundle would fall under this as well.

draconivis (approved):
Awesome, thank you!

PR #446, opened 2026-09-07
Reset the built form cache between requests

ahmed-bhs:
We hit this on a production site running FrankenPHP in worker mode: a visitor opened the contact page and found someone else's name, email and message already filled in.

`Builder` caches every form it builds. The problem is that `buildForm()` runs `handleRequest()` before caching, so what gets stored is a form that already holds whatever was submitted:

```php
$form = $this->createForm(...);
$form->handleRequest($request);   // submitted data lands here
return $form;                      // and this is what gets cached
```

The cache key is `id__type__typeId__locale__name` — no session, no user. So everyone hitting the same form in the same locale gets the same entry.

On php-fpm you never see it, since the container dies with the request. On a worker runtime it stays alive and the next visitor gets the previous one's form.

The fix is a `kernel.reset` tag plus `ResetInterface`, so the container clears the cache between requests. No-op under php-fpm. The cache still does its job within a single request.

Worth noting: `sulu_form.request_listener` already has this tag. The builder just got missed.

## Repro

1. Sulu site with a dynamic form, FrankenPHP `--worker`
2. Visitor A submits
3. Visitor B opens the same page in another browser
4. B sees A's data

## Tests

`Tests/Unit/Form/BuilderTest.php`. Without the fix:

```
Error: Call to undefined method Builder::reset()
Tests: 3, Assertions: 2, Errors: 1, Failures: 1.
```

With it: `OK (3 tests, 3 assertions)`.

Targeting 3.0, happy to port elsewhere if you want it on other branches.

### DIFF
--- a/Form/Builder.php
+++ b/Form/Builder.php
@@ -25,11 +25,12 @@ use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\HttpKernel\Exception\HttpException;
 use Symfony\Component\Security\Csrf\CsrfTokenManagerInterface;
+use Symfony\Contracts\Service\ResetInterface;
 
 /**
  * Builds a dynamic form.
  */
-class Builder implements BuilderInterface
+class Builder implements BuilderInterface, ResetInterface
 {
     /**
      * @var FormInterface[]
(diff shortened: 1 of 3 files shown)


alexander-schranz:
@ahmed-bhs thank you, Nice finding, can you rebase this to 2.6 branch.

ahmed-bhs:
Done, rebased onto 2.6. Could you please tell me if the fix will be also merged on 3.x ?

alexander-schranz:
We merged the service definition changes for current 2.6 which make backporting 2.6 -> 3.0 easier. Can you rebase again and add your changes to the new `.php` based service file.

Fixes will be backmerged to 3.0 by us manually latest before new releases.

ahmed-bhs:
Rebased onto 2.6 — the tag is now in `services.php`. Copilot's reflection note is fixed too.

draconivis (approved):
Thank You! LGTM

Form/Builder.php
```diff
@@ -145,6 +146,23 @@ public function buildByRequest(Request $request): ?FormInterface
...
 
+    /**
+     * Drops the cache of already built forms.
+     *
+     * The cached forms have been through handleRequest(), so they carry the
+     * data submitted by whoever triggered the build. On a persistent runtime
+     * -- FrankenPHP worker mode, RoadRunner, Swoole -- this service outlives
+     * the request, and without this reset the next visitor would be served
+     * the previous visitor's form, already filled in.
+     */
```
Prokyonn:
```suggestion
```

PR #448, opened 2026-09-11
Dynamic::getData(): Return value must be of type array, null returned

LucaMAa:
| Q | A
| --- | ---
| Sulu Version | 3.0.9
| PHP Version | 8,4
| DB Version |  8.4
| Browser Version | Chrome 152.0.7977.83

# Actual Behavior
In the "Data" section of the backoffice for a form containing all the inserted blocks, this error appears:
Dynamic::getData(): Return value must be of type array, null returned
DynamicController::cgetAction

# Expected Behavior

The call correctly returns the data from the form that was just filled out.
# Steps to Reproduce

Create a form using all the blocks.
Fill out the form and submit it.
Go to the "Data" section in the form's back office.

# Possible Solutions
Change This:
```
/**
     * @return mixed[]
     */
    public function getData(): array
    {
        return \json_decode($this->data ?: '[]', true);
    }
```
Into this:
```
/**
     * @return mixed[]
     */
    public function getData(): array
    {
        return \json_decode($this->data ?: '[]', true) ?? [];
    }
```


draconivis:
Thanks @LucaMAa! I was not able to reproduce this with the steps you gave.

I could only trigger your error when that column has an older shape (`text` and/or `utf8mb3` - current installs have `longtext`/`utf8mb4`) and the DB runs a non-strict `sql_mode`, so MySQL silently truncates a write into broken JSON instead of rejecting it.

Could you maybe check with `SELECT @@sql_mode;` and `SHOW FULL COLUMNS FROM fo_dynamics LIKE 'data';`? That could tell us if you're hitting one of those cases.

LucaMAa:
> Thanks @LucaMAa! I was not able to reproduce this with the steps you gave.
> 
> I could only trigger your error when that column has an older shape (`text` and/or `utf8mb3` - current installs have `longtext`/`utf8mb4`) and the DB runs a non-strict `sql_mode`, so MySQL silently truncates a write into broken JSON instead of rejecting it.
> 
> Could you maybe check with `SELECT @@sql_mode;` and `SHOW FULL COLUMNS FROM fo_dynamics LIKE 'data';`? That could tell us if you're hitting one of those cases.

Thanks for checking! Here are the results:

SHOW FULL COLUMNS FROM fo_dynamics LIKE 'data';

data | longtext | utf8mb3_unicode_ci | YES | | | | select,insert,update,references

SELECT @@sql_mode;

IGNORE_SPACE

So it looks like I'm in a mixed state: the column type is already longtext (updated), but the collation is still the old utf8mb3_unicode_ci, and my sql_mode doesn't include STRICT_TRANS_TABLES/STRICT_ALL_TABLES. Sounds like this matches the scenario you described. Let me know if you need anything else to confirm, or if there's a migration/fix I should run on my end.

draconivis:
@LucaMAa thanks for checking! Since this issue came up because of broken data, there is probably no way to recover that data and this error is an indicator for that issue, so falling back to an empty array here is probably not the best solution.

You can try to find the broken rows with this query for example
```sql
SELECT id, formId, LENGTH(data) AS len, LEFT(data, 60) AS head, RIGHT(data, 40) AS tail
FROM fo_dynamics
WHERE data IS NOT NULL AND data <> '' AND data <> '0' AND JSON_VALID(data) = 0;
```

The best solution would probably be to update the database so it does not write invalid JSON by moving the db to the `utf8mb4` character set and also set the sql mode to `strict`.

PR #449, opened 2026-09-14
Use the link provider for external redirect links

### LINKED ISSUE #408: Use Preload function for external link provider
This part should be removed in the future, because there will be a ExternalLinkProvider with a preload function.
See: https://github.com/sulu/sulu/pull/8044

```
        if ('external' === $redirectLink['provider']) {
            return $redirectLink['href'];
        }

```
https://github.com/sulu/SuluFormBundle/blob/2.5/Event/SendinblueListSubscriber.php#L162

Amoifr:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #408
| Related issues/PRs | sulu/SuluHeadlessBundle#148
| License | MIT

#### What's in this PR?

`getUrlFromLink()` returned the raw href as soon as the provider was `external`. That shortcut predates `ExternalLinkProvider` having a `preload()` method, which it now has, so the special case is gone and every provider goes through the pool.

The `external` shortcut is kept for the case where no pool is injected: `Resources/config/type_brevo.php` passes `sulu_markup.link_tag.provider_pool` with `NULL_ON_INVALID_REFERENCE`, so a null pool is a supported state, and falling through to `return null` there would have broken external links.

Two tests come with it. The first checks that an external link goes through the provider and carries its query and anchor, and it fails without the change. The second pins the behaviour when no pool is injected, and passes either way on purpose, as a guard.

#### Why?

An external redirect link dropped its `query` and `anchor`, which every other provider already appended. `ExternalLinkProvider::preload()` returns `new LinkItem($href, '', $href, true)`, so the base URL itself is unchanged and only the missing parts are added.

The `Unit` suite (59 tests) and PHPStan are green locally. The functional suite and the container lints were not run here.


Amoifr:
The workflow run is waiting for a maintainer approval, so I ran the suite locally in the meantime, on PHP 8.5 against MySQL 8.0: Unit is 59 tests / 175 assertions and Functional is 49 tests / 742 assertions, all green.

I also made sure the new test really fails on `3.0` without the change: it produces `https://example.org/page` instead of `https://example.org/page?foo=bar#section`, so the query string and the anchor were the parts getting dropped along the way.

One honest caveat: this machine has no `pdo_pgsql`, so the PostgreSQL part of the matrix is the one thing I could not reproduce. No rush at all, and thanks for maintaining this bundle!

draconivis (approved):
Thank you @Amoifr! This looks already good. Since this is also happening on 2.6, could you please open a PR for that as well? In 2.6 it's the `SendinblueListSubscriber`.

Please rebase this branch as well.

